PR #4, opened 2019-07-02
test users endpoint

tobiasschweizer:


andreas-aeschlimann:
Why have you added @types/node as a dependency? Isn't that a dev dependency?

tobiasschweizer:
> Why have you added @types/node as a dependency? Isn't that a dev dependency?

I moved it to dev deps in https://github.com/dhlab-basel/knora-api-js-lib/pull/4/commits/0a281474fdbe1c7385cb23998954d8862a00d591

PR #7, opened 2019-07-24
Ontologies endpoint

lrosenth:
This PR adds the v2 ontologies endpoint.


tobiasschweizer:
for the integration issue of jsonld see https://github.com/digitalbazaar/jsonld.js/issues/252

tobiasschweizer:
@lrosenth I would like to merge this PR. Could we have a look at it?

I would then like to proceed with parsing resource instances, most probably there will be changes to the ontology reader because of that. This is why I kept the tests basic for now.

PR #9, opened 2019-07-29
feature (cache): add class for caching

### LINKED ISSUE #6: Caching
For ontology information and lists, we have to provide caches that also work if there are async   requests for the same entity.

tobiasschweizer:
This PR adds a generic caching class using generics.

resolves #6


tobiasschweizer:
@andreas-aeschlimann Your proposal works! 

https://github.com/dhlab-basel/knora-api-js-lib/blob/486c13726b860b3832e67227544f700fe0e7b2b1/src/cache/GenericCache.ts#L24-L44

The cache returns an  `AysncSubject` for an entry. If the entry does not exists yet, an "empty" `AsyncSubject` is created in the cache and returned. Once the answer from Knora is available, the `AsyncSubject` is updated and completed.

I added tests and also tried it in https://github.com/dhlab-basel/knora-api-js-lib-test

I think the `AsyncSubject` is preferable to the `BehaviorSubject` in this case because there is not initial value needed and once the value is there it is completed.

tobiasschweizer:
idea to load additional items:

```ts
this.requestItemFromKnora(key).subscribe((response: T) => {
            this.cache[key].next(response);

            // TODO: ask for items (`this.getItem(key)`) the current item depends on (array of AsyncSubject) and subscribe to forkJoin, calling `this.cache[key].complete()`

            // mock 
            const dependencies = [new AsyncSubject(), new AsyncSubject()];

            dependencies.forEach((item: AsyncSubject<number>, index: number) => {
                item.next(index);
                console.log(index);
                item.complete();
            });

            forkJoin(dependencies).subscribe(() => {
                console.log('completing ' + key);
                this.cache[key].complete();
            });
        });
```

tobiasschweizer:
# Note on Dependencies:

When an item is requested from the cache, all of its dependencies must have be resolved when the requested item is returned from the cache (the item itself and all of its dependencies are present in the cache). This is implemented using `forkJoin`: once all dependencies of an item are ready, the current item itself can be completed.

`getItem(key: string, requestedDeps: string[] = [])` calls itself recursively to resolve a dependency, passing along the keys of the items that are already being processed (completed or not). The keys of those items are ignored when looking at dependencies of a newly obatinaed item. This solves the problem of stalling processes waiting for each other to complete (1 depends on 2 and vice versa).

However, I am not sure if this is sufficient when there are multiple asynchronous processes with shared dependencies. Could there be a case in which a dependency "a" is required for the whole process to complete, but a relies on "b" which itself is waiting for "a", resulting a stall?

tobiasschweizer:
I have taken the following decision: Dependencies are resolved asynchronously. Once the requested item completes, its dependencies might not have been resolved yet, however, they have been requested.

I have taken this decision because I want to avoid a situation of stalling requests, i.e. a request that cannot complete because another request has not completed yet (see https://github.com/dhlab-basel/knora-api-js-lib/pull/9#issuecomment-519588419). With the current design, a request completes as soon as the information has been fetched from Knora. If it has dependencies, these are requested too without waiting for them.

lrosenth:
Seems to be OK and very reasonable. Thanks

lrosenth:
Seems to be OK and very reasonablee

PR #14, opened 2019-08-26
Wip/parse resources

### LINKED ISSUE #17: check for forkJoin on empty arrays
https://stackoverflow.com/questions/53477784/rxjs-forkjoin-with-empty-array?rq=1

https://www.learnrxjs.io/operators/conditional/defaultifempty.html

tobiasschweizer:
This PR adds functionality to parse resource responses serialized as JSON-LD.

closes #17


src/api/v2/resource/resources-endpoint.ts
```diff
@@ -0,0 +1,164 @@
...
+
+                const colorVal: ReadColorValue = this.jsonConvert.deserialize(valueJsonld, ReadColorValue) as ReadColorValue;
+                value = of(colorVal);
+                break;
+            }
+
+            case Constants.DateValue: {
+
+                const dateVal: ReadDateValue = this.jsonConvert.deserialize(valueJsonld, ReadDateValue) as ReadDateValue;
+                dateVal['parseDate']();
```
tobiasschweizer: @andreas-aeschlimann Is there a better way to do this? In a first step, the JSON-LD props are mapped to the class's members. But for more convenience, the data needs to be converted to a date class to handle periods and precisions. Then the client wouldn't actually have to see what originally was submitted in the JSON-LD.



```ts
@JsonObject('ReadDateValue')
	export class ReadDateValue extends ReadValue {
	
	    date: Date | Period;
	
	    @JsonProperty(Constants.DateValueHasCalendar, String)
	    private calendar: string = "";
	
	    @JsonProperty(Constants.DateValueHasStartDay, Number, true)
	    private startDay?: number = undefined;
	
	    @JsonProperty(Constants.DateValueHasStartMonth, Number, true)
	    private startMonth?: number = undefined;
	
	    @JsonProperty(Constants.DateValueHasStartYear, Number)
	    private startYear: number = 0;
	
	    @JsonProperty(Constants.DateValueHasStartEra, String)
	    private startEra: string = "";
	
	    @JsonProperty(Constants.DateValueHasEndDay, Number, true)
	    private endDay?: number = undefined;
	
	    @JsonProperty(Constants.DateValueHasEndMonth, Number)
	    private endMonth?: number = undefined;
	
	    @JsonProperty(Constants.DateValueHasEndYear, Number)
	    private endYear: number = 0;
	
	    @JsonProperty(Constants.DateValueHasEndEra, String)
	    private endEra: string = "";
	
	    private parseDate() {
	
	        if (this.startYear === this.endYear && this.startMonth === this.endMonth && this.startDay === this.endDay && this.startEra === this.endEra) {
	            // precise date
	            this.date = new Date(this.calendar, this.startEra, this.startYear, this.startMonth, this.startDay);
	        } else {
	            // date period
	            this.date = new Period(new Date(this.calendar, this.startEra, this.startYear, this.startMonth, this.startDay), new Date(this.calendar, this.endEra, this.endYear, this.endMonth, this.endDay));
	        }
	    }
	}
```

tobiasschweizer: @andreas-aeschlimann Do you think I should make another class with a constructor that could be instantiated once the JSON was successfully deserialized?

src/models/v2/resources/values/read-date-value.ts
```diff
@@ -0,0 +1,101 @@
...
+export enum Precision {
+    yearPrecision,
+    monthPrecision,
+    dayPrecision
+}
+
+/**
+ * Represents a Salsah date object with a precision information.
+ */
+export class Date {
```
andreas-aeschlimann: I suggest renaming this to SalsahDate (and also use SalsahPeriod) because this could lead to confusion for a developer in comparison to the JavaScript Date object?

tobiasschweizer: Good point! I will do that.

src/api/v2/resource/resources-endpoint.ts
andreas-aeschlimann: I don't fully understand the purpose of this class yet. Why not create a custom converter that contains the `parseDate()` functionality? Then you probably do not even need this class and you can directly use `Date | Period` instead of `ReadDateValue` where you need it.

src/models/v2/resources/ResourcesConversionUtil.ts
```diff
@@ -0,0 +1,320 @@
...
+            case Constants.ColorValue: {
+                value = handleSimpleValue(valueJsonld, ReadColorValue, jsonConvert);
+                break;
+            }
+
+            case Constants.DateValue: {
+                const dateVal = handleSimpleValue(valueJsonld, ReadDateValue, jsonConvert) as Observable<ReadDateValue>;
+                value = dateVal.pipe(map(
+                    date => {
+                        date["parseDate"]();
```
andreas-aeschlimann: I would rather have a method public than calling it like that. The compiler will not notice any changes afaik and not complain if there is any change.

andreas-aeschlimann:
What about the 1 class per file convention? 🤐

(It is not priority of course, but as the project is growing, it gets more work to fix.)

src/models/v2/resources/ResourcesConversionUtil.ts
tobiasschweizer: I agree that there should be a better way to do this. This is related to https://github.com/dhlab-basel/knora-api-js-lib/pull/14#discussion_r322204391

src/api/v2/resource/resources-endpoint.ts
tobiasschweizer: I would prefer a second class that can be instantiated once the JSON has been successfully parsed. And this second class has a constructor which takes arguments from the instance that json2typescript created.

src/models/v2/resources/values/read-date-value.ts
tobiasschweizer: done in https://github.com/dhlab-basel/knora-api-js-lib/pull/14/commits/27bd6e18f7cc4083e4e4bc58bf6e61653ec49696

I used "Knora" as a prefix though.

src/api/v2/resource/resources-endpoint.ts
tobiasschweizer: an implementation is provided in https://github.com/dhlab-basel/knora-api-js-lib/pull/14/commits/1161c6f14319d5f1ac18324f3e2411ccf3cf5b0f

src/models/v2/resources/values/read-date-value.ts
tobiasschweizer: it's specific to Knora, not SALSAH

tobiasschweizer:
@andreas-aeschlimann Could you pleas approve this PR?

PR #22, opened 2019-09-09
Generic cache: unsubscribe automatically 

### LINKED ISSUE #21: Automatically unsubscribe in GenericCache
Use `take` operator to automatically unsubscribe when first value has been emitted:

```ts

 this.requestItemFromKnora(key).pipe(take(1)).subscribe(...)

```

We expect exactly one value. 

https://blog.angularindepth.com/the-best-way-to-unsubscribe-rxjs-observable-in-the-angular-applications-d8f9aa42f6a0

If an error occurs, `AsyncSubject` should probably fail too. Maybe add error callback and let `AsyncSubject` fail in that case.

tobiasschweizer:
This PR makes `GenericCache` unsunscribe automatically once the requested value is received from Knora.

closes #21

### DIFF
--- a/src/cache/GenericCache.ts
+++ b/src/cache/GenericCache.ts
@@ -1,4 +1,5 @@
 import { AsyncSubject, Observable } from "rxjs";
+import { take } from "rxjs/operators";
 
 /**
  * Generic cache class.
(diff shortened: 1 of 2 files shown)


tobiasschweizer:
@andreas-aeschlimann Do we have to call `error` on the `AsyncSubject` if the request to Knora fails? So that the subscriber to the cache gets an appropriate error message.

PR #29, opened 2019-09-10
Transform test data

tobiasschweizer:
This PR automatically expands the prefixes of JSONLD. Like this, mocked services can use this JSONLD directly without having to deal with the extra complexity of using the jsonld processor.


tobiasschweizer:
This way, the test data provided by Knora can be automatically transformed, rather than having to do this manually.

PR #38, opened 2019-09-13
Wip/19 improve jsonld import

### LINKED ISSUE #19: Improve jsonld import
In OntologiesEndpoint, we should improve the import of the JSON LD library.

```
const jsonld = require("jsonld/dist/jsonld.js");
```

andreas-aeschlimann:
This pull requests attempts to fix the issue with the json ld import and would close #19.

While I was able to fix some ES2017 issues, I still don't get the tests to work because of the spread operator.

This issue helped reducing the compile errors, but the spread operator still causes a failure: https://github.com/monounity/karma-typescript/issues/344

I have set a bounty on an older Stackoverflow question about a similar issue: https://stackoverflow.com/questions/53412866/karma-typescript-import-js-file-with-async-keyword


andreas-aeschlimann:
If you find an answer, you can fix this @tobiasschweizer. I don't know.

I have seen that you were suggested to use jest. I am against using anything related to FB, but we also can give it a try if we fail with karma.

tobiasschweizer:
I don't think we can use a version containing ES2017 code unless this is going to be transpiled to ES5 before being executed. Also if we manage to make the tests work in karma, what about the browsers? The will be ES2015 compatible at most.

But I think this should be handled by the provider of the npm package (see https://github.com/digitalbazaar/jsonld.js/issues/252#issuecomment-515885921).

PR #39, opened 2019-09-13
Wip/35 make jsonconvert static

andreas-aeschlimann:
This PR closes #35.

I moved the `JsonConvert` instance to `KnoraApiConfig` because it makes more sense there as a global static config.

It is questionable though if we still need to pass `KnoraApiConfig.jsonConvert` as parameter through all these functions, because `jsonConvert` could be directly accessed. This could be added in another commit.


andreas-aeschlimann:
What do you think @tobiasschweizer?

tobiasschweizer:
> It is questionable though if we still need to pass KnoraApiConfig.jsonConvert as parameter through all these functions, because jsonConvert could be directly accessed. This could be added in another commit.

I think we could access it directly. Could you make an issue for that so it can be done it a subsequent PR?

tobiasschweizer:
This PR requires the generated code from Knora to be changed as well. I propose to implement a solution that requires the generated code to make as few assumptions as possible about how the result is going to be processed, see https://github.com/dhlab-basel/Knora/pull/1392#issuecomment-521276363.

PR #44, opened 2019-09-17
feature (ontology): add missing properties

### LINKED ISSUE #43: Add missing properties to Ontology Entities
Properties (`src/models/v2/ontologies/resource-property-definition.ts`):
- ~~guiorder (along with cardinality)~~
- ~~isEditable~~,
- ~~isLinkProperty~~,
- ~~isLinkValueProperty~~
- ~~guiAttribute~~
- ~~isInherited~~

Resource classes (`src/models/v2/ontologies/resource-class-definition.ts`):
- ~~canBeInstantiated~~

tobiasschweizer:
This PR adds missing properties to property and resource class defs.

closes #43


src/models/v2/custom-converters/gui-attribute-converter.ts
```diff
@@ -0,0 +1,21 @@
+import { JsonConverter, JsonCustomConvert } from "json2typescript";
+
+@JsonConverter
+export class GuiAttributeConverter implements JsonCustomConvert<string[]> {
+
+    serialize(attrs: string[]): any {
+    }
```
andreas-aeschlimann: Should we throw exceptions for the missing serializations? I know there are some more, just asking in general.

tobiasschweizer: I think we should think about how to create or modify ontologies / resources / values as soon as possible. In these cases, we would need the `serialize` method.

PR #46, opened 2019-09-18
Specify value type of property

tobiasschweizer:
implements #54
closes #45

PR #54, opened 2019-09-20
Feature: added utility function to get type of value

lrosenth:
Added utility function ```getValuesType(propname)```to read-resources.ts which returns the data type of the value(s).
Added a test to ```resourcesConversionUtil.spec.ts```...

### DIFF
--- a/src/models/v2/resources/read-resource.ts
+++ b/src/models/v2/resources/read-resource.ts
@@ -63,6 +63,14 @@ export class ReadResource {
         }
     }
 
+    getValuesType(property: string): string {
+        if (this.properties.hasOwnProperty(property) && (this.properties[property].length > 0)) {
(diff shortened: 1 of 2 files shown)


src/models/v2/resources/read-resource.ts
```diff
@@ -63,6 +63,14 @@ export class ReadResource {
         }
     }
```
tobiasschweizer: We could also get this information from the ontology, so we could also return a type if there is no instance of a value.

tobiasschweizer:
I will implement the proposed feature in #46 46

tobiasschweizer:
done in https://github.com/dhlab-basel/knora-api-js-lib/pull/46/commits/c4ae2e8fff35e9511b5a49fc41476a7298ddcedd

Issue #41, opened 2019-09-16
Support knora-api:timeValue


tobiasschweizer:
implement support for https://github.com/dhlab-basel/Knora/pull/1403

PR #56, opened 2019-09-23
Integrate v2 test data

tobiasschweizer:
This PR adds a script to automate the integration of v2 test data.

Needed for #47 and #41


tobiasschweizer:
@benjamingeer This is a simple way to keep the test data in sync. It just copies JSON test data files from the Knora v2 E2E tests. I would like all tests to be based on the anything project. I would need to produce more test data for anything, like ontologies, lists etc. 

I could just add some more E2E tests to Knora. Would that be ok?

tobiasschweizer:
> Would it be more maintainable if it had a config file specifying the source and destination paths?

That would be the first step, probably we want to have a process like for the admin API with a special route.

tobiasschweizer:
@benjamingeer I think I could rewrite that script in JS to be run by node. The config could live in a JSOn file.

benjamingeer:
I think you'll be happier with it in Node (or Python) than in Bash.

tobiasschweizer:
> I think you'll be happier with it in Node (or Python) than in Bash.

I think the effect on my mental health will be minimal ;-)

tobiasschweizer:
done in https://github.com/dhlab-basel/knora-api-js-lib/pull/56/commits/cc333efc0f299124619ad94a2498ea71211a8320 (please note the commit message)

PR #74, opened 2019-10-16
Wip/45 value type

lrosenth:
Added a generic member variable "strval" to ReadValue in order to have easy access to a string representation of all value types

Added extensive tests


tobiasschweizer:
@lrosenth I could get rid of the casts and  `unknown` since the type can be determined by the transpiler.

src/models/v2/resources/read-resource.ts
```diff
@@ -83,6 +83,15 @@ export class ReadResource {
         }
     }
 
+    getValuesAsStringArray(property: string, defstr: string = '?'): string[] {
```
tobiasschweizer: @lrosenth What is the purpose of `defstr`?

PR #108, opened 2019-11-19
fix(admin): Make all properties in UpdateProjectRequest optional

### LINKED ISSUE #100: Optional properties in UpdateProjectRequest
Can we set all properties in UpdateProjectRequest class as optional. In my case I have to update project status only (restore deleted project) but keywords and description are needed. Or is there a reason for that?

benjamingeer:
Fixes #100.

Needs https://github.com/dasch-swiss/knora-api/pull/1519.


benjamingeer:
Thanks for the review!

PR #114, opened 2019-12-02
Additional caches

### LINKED ISSUE #106: Expand cache services
Can you expand the cache service with the following methods:

- get project by shortcode
- get project members by shortcode
- get project groups by shortcode
- get all users
- get user by username
- get ontology by iri

tobiasschweizer:
- [x] get user by username
- [x] get project by shortcode
- [ ] get project members by shortcode
- [ ] get groups 
- [ ] get users

closes #106


tobiasschweizer:
@kilchenmann I have just realized that the generic cache I wrote does not work for getting project members by project shortcode because the `MembersResponse` does not have an id.

Also I have realized that the generic cache requires an item to be identified by an id, but users and projects can be identified in three different ways. Of course we could a separate cache for each way of getting a user, but there would be redundancy.

kilchenmann:
> I have just realized that the generic cache I wrote does not work for getting project members by project shortcode because the MembersResponse does not have an id.

I handled it with a key for the cache like `'members_of_' + this.projectcode`. But it's ok. I will optimize and use my generic cache service in the app itself.

PR #115, opened 2019-12-03
add method to create a resource

tobiasschweizer:
Add method to create a resource.


benjamingeer:
OK, will do.

tobiasschweizer:
note to myself: understand JS number precision https://stackoverflow.com/questions/1458633/how-to-deal-with-floating-point-number-precision-in-javascript

tobiasschweizer:
@benjamingeer I had to change "100000000000000.000000000000001" to 1.5

benjamingeer:
> I had to change "100000000000000.000000000000001" to 1.5

https://www.npmjs.com/package/js-big-decimal

tobiasschweizer:
Yes, I saw his too (https://github.com/dasch-swiss/knora-api/pull/1549#issuecomment-564599052). Yet another dependency ...

benjamingeer:
> Yes, I saw his too (dasch-swiss/knora-api#1549 (comment)). Yet another dependency ...

Or you can just use `string`. But we can't lose precision, that would be equivalent to losing data.

tobiasschweizer:
> Or you can just use string. But we can't lose precision, that would be equivalent to losing data.

I will look at this and try to figure something out. I remember that we had this problem in Scala too before using `BigDecimal`.

tobiasschweizer:
@benjamingeer I think I will create a separate issue to deal with precisions

PR #128, opened 2019-12-16
Admin List

tobiasschweizer:
Integrates https://github.com/dasch-swiss/knora-api/pull/1509

PR #129, opened 2019-12-16
Add cache for list admin

tobiasschweizer:
needs #128


tobiasschweizer:
@benjamingeer I would like to instantiate the cache in the `AdminEndpoint` but this file is generated by Knora. Could I simply add the necessary statement in the template? But then this would only compile if the `ListAdminCache` exists.

tobiasschweizer:
@lrosenth I will add the cache for admin lists here (ready once #128 has been merged).

tobiasschweizer:
@lrosenth I added the cache you wrote here. Please not that the `AdminEndpoint` does not instantiate it.

benjamingeer:
> Could I simply add the necessary statement in the template? But then this would only compile if the ListAdminCache exists.

`AdminEndpoint` is generated using the template `generateTypeScriptMainEndpoint.scala.txt`, which could also be used for generating `V2Endpoint` in the future. So we have to keep it generic. How about adding it to `Endpoint`?

tobiasschweizer:
> How about adding it to Endpoint?

The caches are either bound to the v2 or admin endpoint.

https://github.com/dasch-swiss/knora-api-js-lib/blob/f48e16faa3fc578b4b3062e6467dc29d437700c2/src/api/v2/v2-endpoint.ts#L60-L61

For instance, the `OntologyCache` needs access to the v2 endpoint to fetch data from Knora, if not already in the cache:

https://github.com/dasch-swiss/knora-api-js-lib/blob/f48e16faa3fc578b4b3062e6467dc29d437700c2/src/cache/OntologyCache.ts#L33-L35

https://github.com/dasch-swiss/knora-api-js-lib/blob/f48e16faa3fc578b4b3062e6467dc29d437700c2/src/cache/OntologyCache.ts#L144-L148

benjamingeer:
But why not add all the caches to the base class `Endpoint`, and let the subclasses use whichever ones they need?

benjamingeer:
Oh I see, the caches need to know about the sub-endpoints... Hmm...

tobiasschweizer:
> Oh I see, the caches need to know about the sub-endpoints... Hmm...

The caches are an additional layer to the endpoint and decide whether the endpoint has to be used (call to Knora) or an information is already in the cache.

tobiasschweizer:
I think this is really a thing that belongs to the design of the client lib and I don't think it should be generated by Knora.

benjamingeer:
> I think this is really a thing that belongs to the design of the client lib and I don't think it should be generated by Knora.

The problem is that if each cache is specific to a particular type of data, it needs to know about generated classes and generated functions. This means that there is one more thing to break when those classes and functions change.

And if the endpoint needs to instantiate the cache, then something in the code generation framework needs to know that this endpoint requires this cache.

Here's an idea: if we want people to use a cache, maybe it doesn't make sense to provide a generated function (like `getList` or `getOntology`) that don't use the cache. Wouldn't it be better if the generated function called the cache? Then it would be impossible not to use the cache.

Could we have just one generic cache class, which would cache any type of object (it would just be a simple key-value store)? Then:

- the client API definition in Scala would specify which functions need to use a cache
- the template that generates the endpoint would:
    - instantiate a cache instance for each of those functions
    - generate something in each function to call the cache

tobiasschweizer:
I will have to check with @rosenth how we want to proceed here. He needs a new version of @knora/api today, and I will have to work on its integration into Knora-UI.

benjamingeer:
OK. I'll be on holiday starting tomorrow, so if you want to merge #1509, you'll have to do it without me or wait till January.

subotic:
In general I would be rather conservative with caching in the client, since cache invalidation is not trivial and always lead to an increase in complexity. Also, the complete admin API will be cached on the server some day (some routes are already). The client should only cache things that provide a substantial performance increase in the client while accepting the associated cost of stale data.

Further, the way caching is employed should be left to the client implementation, which would mean that there should be a cached and non-cached methods in the client library ...

PR #142, opened 2020-01-28
Util for permission handling

### LINKED ISSUE #140: Parse Permissions
There should be a method that takes care of parsing permission strings (https://docs.knora.org/paradox/02-knora-ontologies/knora-base.html#permissions). Also the method should handle the hierarchy of the different permissions: higher permissions imply lesser permissions (https://docs.knora.org/paradox/03-apis/api-v2/reading-user-permissions.html).

Idea: the single permissions could be represented by an enumeration. When reading permissions, the given permissions and the lesser permissions could be added too an array. Then the client could simply check for the presence of a specific permissions and would not have to be aware of the hierarchical relation of permissions.

tobiasschweizer:
This PR adds a util that handles permission codes returned by Knora.

closes #140


tobiasschweizer:
note to myself: add it to index.ts

PR #147, opened 2020-02-10
tests (it): adapt assertions

tobiasschweizer:
This PR adapts assertions since Knora test data has changed.


tobiasschweizer:
@subotic Test data has changed in Knora (sample searches returns more results than before) and the integration tests has to be adapted accordingly.

PR #163, opened 2020-03-13
Determine link prop Iri form link value prop…

### LINKED ISSUE #162: Make Helper Function to determine a link prop from a link value prop
Given `anything:hasOtherThingValue` the method should return `hasOtherThing`.

tobiasschweizer:
closes #162

### DIFF
--- a/src/models/v2/resources/read/read-resource.ts
+++ b/src/models/v2/resources/read/read-resource.ts
@@ -1,4 +1,5 @@
 import { JsonObject, JsonProperty } from "json2typescript";
+import { ResourcePropertyDefinition } from "../../../..";
 import { IResourceClassAndPropertyDefinitions } from "../../../../cache/OntologyCache";
 import { Constants } from "../../Constants";
 import { DateTimeStampConverter } from "../../custom-converters/date-time-stamp-converter";
@@ -93,9 +94,20 @@ export class ReadResource extends ReadWriteResource {
      * @param linkValueProperty IRI of the link value property.
      */
     getLinkPropertyIriFromLinkValuePropertyIri(linkValueProperty: string) {
-        if (linkValueProperty.endsWith("Value")) {
+        if (this.entityInfo.properties[linkValueProperty] !== undefined
+            && this.entityInfo.properties[linkValueProperty] instanceof ResourcePropertyDefinition
+            && (this.entityInfo.properties[linkValueProperty] as ResourcePropertyDefinition).isLinkValueProperty
+            && linkValueProperty.endsWith("Value")) {
             // remove "Value" from the end of the string and return the Iri
-            return linkValueProperty.substring(0, linkValueProperty.length - 5);
+            const linkPropIri = linkValueProperty.substring(0, linkValueProperty.length - 5);
+
+            if (this.entityInfo.properties[linkPropIri] !== undefined
+                && this.entityInfo.properties[linkPropIri] instanceof ResourcePropertyDefinition
+                && (this.entityInfo.properties[linkPropIri] as ResourcePropertyDefinition).isLinkProperty) {
+                return linkPropIri;
+            } else {
+                return new Error(`Could not determine link property IRI for ${linkValueProperty}`);
+            }
         } else {
             throw new Error(`${linkValueProperty} is not a valid link value property IRI`);
         }
--- a/src/models/v2/resources/read/read-resource.ts
+++ b/src/models/v2/resources/read/read-resource.ts
@@ -87,6 +87,21 @@ export class ReadResource extends ReadWriteResource {
     }
 
     /**
+     * Given the IRI of a property pointing to a link value,
+     * returns the corresponding link property IRI.
+     *
+     * @param linkValueProperty IRI of the link value property.
+     */
+    getLinkPropertyIriFromLinkValuePropertyIri(linkValueProperty: string) {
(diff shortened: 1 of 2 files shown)


src/models/v2/resources/read/read-resource.ts
```diff
@@ -86,6 +86,21 @@ export class ReadResource extends ReadWriteResource {
         }
     }
 
+    /**
+     * Given the IRI of a property pointing to a link value,
+     * returns the corresponding link property IRI.
+     *
+     * @param linkValueProperty IRI of the link value property.
+     */
+    getLinkPropertyIriFromLinkValuePropertyIri(linkValueProperty: string) {
```
SepidehAlassi: @tobiasschweizer As I mentioned yesterday, I believe this type of regex is not the solution. What if a user has ended the property name also with Value. eg: property: "somethingValue" then LinkValue would be "somethingValueValue". This code would trim the "Value" from both property and its linkValue. That would be actually wrong. I believe the API internally defines the connection between a property and its linkValue. In js-lib we should exploit that connection instead of applying regex.

tobiasschweizer: I think there is just a naming convention that a link value property ends with "Value" and the corresponding link property is the same just without "Value" at the end. @benjamingeer Is that correct?

benjamingeer: Yes:

> The name of the link value property is constructed using a simple naming convention:
> the word `Value` is appended to the name of the link property.

https://docs.knora.org/paradox/02-knora-ontologies/knora-base.html#links-between-resources

In Knora, the `SmartIri` class has methods `fromLinkPropToLinkValueProp` and `fromLinkValuePropToLinkProp`, which just add and remove the word `Value`.

benjamingeer: Also, on startup, Knora checks that in each ontology, there is a link value property for each link property and vice versa.

tobiasschweizer: >  I believe the API internally defines the connection between a property and its linkValue. 

I think that would actually make sense, but the relation is not defined explicitly.

tobiasschweizer: @SepidehAlassi Do you think that you can use this in https://github.com/dasch-swiss/knora-ui-ng-lib/pull/27?

benjamingeer: You could find out whether a property is really a link property, or really a link value property, by checking its `knora-api:objectType` in its ontology.

tobiasschweizer: Ok, that would be possible since `ReadResource` contains all the necessary information. I will do that. But still the relation between the two properties in only based on the naming convention.

benjamingeer: > But still the relation between the two properties in only based on the naming convention.

That's right. But I think that if you have property `example:hasBook` with `knora-api:objectType example:Book`, and property `example:hasBookValue` with `knora-api:objectType knora-api:LinkValue`, you can be sure that they go together.

SepidehAlassi: @benjamingeer I was aware of this regulation for the definition of links in the ontology. But I was under the impression that API aligns the link value property to its corresponding link property. Then one could always request the link property of each link value property. In this case, we would not need to check if the property name ends with "Value" or not. Instead, we could just check the ObjectType, if it was linkValue, we would request its corresponding link property.

tobiasschweizer: > You could find out whether a property is really a link property, or really a link value property, by checking its `knora-api:objectType` in its ontology.

done in https://github.com/dasch-swiss/knora-api-js-lib/pull/163/commits/4e0d7d800fc2e8ac17893d7139a18eba0174a282

benjamingeer: Internally, no connection is stored between a link property and the corresponding link value property. We just use the naming convention, and the type information in the property definitions, to convert from one to another and check for errors.

To determine whether a property is a link property or a link value property, Knora doesn't rely on whether the property's name ends in `Value`. Instead, it looks at the property's object type as defined in its ontology, which says whether its object should be a resource or a link value. Then, once Knora is sure which kind of property it is, it adds or removes `Value` as needed to convert to the other kind of property. It then checks that this other property exists as well, etc.

tobiasschweizer:
I think this PR is not the place to have a debate whether the relation between link value and link properties should be represented differently by Knora. Please create an issue in knora-api repo for that.

tobiasschweizer:
I added a test for a non link value prop just to be sure: https://github.com/dasch-swiss/knora-api-js-lib/pull/163/commits/384897912309e35786c4958f9e8b7ef90e4005a0

tobiasschweizer:
@SepidehAlassi Is this PR ok for you?

tobiasschweizer:
@SepidehAlassi Ok thanks, I will merge this PR and update https://github.com/dasch-swiss/knora-api-js-lib/pull/125 (branch we are using in knora-ui-ng-lib tests) as soon as the tests passed. Then you can use it in your PR.

tobiasschweizer:
Merged :-)

tobiasschweizer:
do not forget to rebuild knora-api-js-ib and republish it with yalc :-)

PR #164, opened 2020-03-16
Fix import from index.ts

tobiasschweizer:
This PR fixes an import from `index.ts` that resulted in a circular dep.

### DIFF
--- a/src/models/v2/resources/read/read-resource.ts
+++ b/src/models/v2/resources/read/read-resource.ts
@@ -1,5 +1,4 @@
 import { JsonObject, JsonProperty } from "json2typescript";
-import { ResourcePropertyDefinition } from "../../../..";
 import { IResourceClassAndPropertyDefinitions } from "../../../../cache/OntologyCache";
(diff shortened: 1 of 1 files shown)


tobiasschweizer:
> I was going to write to you about that circular dependency. thanks for fixing it.

My IDE likes to import from `index.ts` although it should not ...

SepidehAlassi:
@tobiasschweizer mine does the same. you also use Webstorm, don't you? I noticed the warnings on the command line.

tobiasschweizer:
> @tobiasschweizer mine does the same. you also use Webstorm, don't you? I noticed the warnings on the command line.

Webstorm offers multiple imports, the one from `index.ts` being the default. You can also choose a different path, but that is not the default behaviour. I wonder if this can be changed.

SepidehAlassi:
I have tried this fix, it works fine. You can merge this PR.

tobiasschweizer:
> I have tried this fix, it works fine. You can merge this PR.

Thx. Just waiting for the tests to pass ...

PR #168, opened 2020-03-17
Fix return type of link prop method

tobiasschweizer:
fixes a problem with a method's return type, see https://github.com/dasch-swiss/knora-ui-ng-lib/pull/27#discussion_r393819787


tobiasschweizer:
> :-D thanks holy noodlie goodness for this fix.

Spaghetti monster? :-)

PR #171, opened 2020-04-08
Test Framework: Do Not Check for Exact Number of Search Results in E2E Tests

tobiasschweizer:
This PR changes the assertions in the E2E tests so that search tests do not expect an exact number of results (see https://github.com/dasch-swiss/knora-api/pull/1615#issuecomment-610252259). The data can always change in knora-api, but this is irrelevant to these tests. They just have to make sure that the result can be obtained and deserialized correctly.

### DIFF
--- a/test-framework/e2e/src/app.e2e-spec.ts
+++ b/test-framework/e2e/src/app.e2e-spec.ts
@@ -1,6 +1,16 @@
 import {AppPage} from './app.po';
 import {browser, logging} from 'protractor';
 
+/**
+ * Attempts to convert a string to a number.
+ *
+ * @param numStr string to be converted to a number.
+ */
+const convertTextToNumber = (numStr): number => {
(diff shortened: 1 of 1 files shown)


tobiasschweizer:
Once this PR has been merged, the JS lib tests will pass for knora-api.

tobiasschweizer:
> but I really think we should remove these tests

you mean we should run them against a specific release of knora-api and remove the setup from github-ci as it is now?

benjamingeer:
> you mean we should run them against a specific release of knora-api and remove the setup from github-ci as it is now?

Yes, exactly.

tobiasschweizer:
> > you mean we should run them against a specific release of knora-api and remove the setup from github-ci as it is now?
> 
> Yes, exactly.

Ok, let's do this once we have decided how to proceed with releases.

PR #176, opened 2020-04-15
Make RxJS a Peer Dep

tobiasschweizer:
This PR makes RxJS a peer dependency of the lib. This is necessary because the lib exposes RxJS's `Observable` in its public API. The lib is used by an Angular application that also requires RxJS. If the two versions are different, then the two Observables are treated as if they were different classes, resulting in a compile error like

> Types of property 'source' are incompatible.
        Type 'import(".../knora-api-js-lib/test-framework/node_modules/@knora/api/node_modules/rxjs/internal/Observable").Observable<any>' is not assignable to type 'import(".../test-framework/node_modules/rxjs/internal/Observable").Observable<any>'.

I think this is exactly what peer dependencies attempt to solve:

> But now what if JacksModule exposes its dependency on JillsModule in some way. It accepts an instance of JillsClass for example... What happens when we create a new JillsClass using version 2.0 of the library and pass it along to jacksFunction? All hell will break loose! Simple things like jillsObject instanceof JillsClass will suddenly return false because jillsObject is actually an instance of another JillsClass, the 2.0 version.

https://stackoverflow.com/questions/26737819/why-use-peer-dependencies-in-npm-for-plugins#answer-34645112

### DIFF
--- a/test-framework/src/app/app.component.ts
+++ b/test-framework/src/app/app.component.ts
@@ -27,6 +27,8 @@ import {
   DeleteResource
 } from '@knora/api';
 
+import {map} from 'rxjs/operators'
+
 @Component({
   selector: 'app-root',
   templateUrl: './app.component.html',
--- a/test-framework/src/app/app.component.ts
+++ b/test-framework/src/app/app.component.ts
@@ -126,8 +126,7 @@ export class AppComponent implements OnInit {
 
     this.knoraApiConnection.v2.res.getResource(iri).pipe(
         map(
-            (res) => {
-                console.log('test');
+            (res) => { // make sure RxJS versions (Observable) are compatible
                 return res;
             })
     ).subscribe(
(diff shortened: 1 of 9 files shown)


lrosenth:
I looked at the changes – seems exactlyto be what we need! Thanks  lot!
Please merge...

tobiasschweizer:
and another very good explanation:

> The Guidelines#
Favor using Peer Dependencies when one of the following is true:

> Having multiple copies of a package would cause conflicts
The dependency is visible in your interface
You want the developer to decide which version to install
Let’s take the example of angular/core. Obviously, if you are creating an Angular Library, angular/core is going to be a very visible part of your library’s interface. Hence, it belongs in your peerDependencies.

> However, maybe your library uses Moment.js internally to process some time related inputs. Moment.js most likely won’t be exposed in the interface of your Angular Services or Components. Hence, it belongs in your dependencies.

https://indepth.dev/npm-peer-dependencies/

tobiasschweizer:
@subotic @lrosenth @andreas-aeschlimann I have also mentioned all used of `RxJS` to facilitate upgrades to future versions that might have breaking changes.

test-framework/src/app/app.component.ts
```diff
@@ -123,7 +124,13 @@ export class AppComponent implements OnInit {
 
   getResource(iri: string) {
 
-    this.knoraApiConnection.v2.res.getResource(iri).subscribe(
+    this.knoraApiConnection.v2.res.getResource(iri).pipe(
+        map(
+            (res) => {
+                console.log('test');
```
mdelez, line 130: please remove test log :)

tobiasschweizer: good catch, will do ;-)

tobiasschweizer: done in https://github.com/dasch-swiss/knora-api-js-lib/pull/176/commits/18ee78111506bc9d2f156fee56b8e57f9d9e299d

PR #179, opened 2020-04-22
Add missing exports

tobiasschweizer:
https://dasch.myjetbrains.com/youtrack/issue/DSP-182


tobiasschweizer:
I cannot export admin lists endpoint because this would conflict with v2 lists endpoint. I think each endpoint should have a v2 or admin suffix.

tobiasschweizer:
@kilchenmann Will this break anything in knora-app?

kilchenmann:
> @kilchenmann Will this break anything in knora-app?

yes of course. Why it shouldn't?

PR #181, opened 2020-05-06
Prepare Dev Package

tobiasschweizer:
This PR introduces a script that prepares a dev package for publication:

- add files from "test" in `package.json`
- add exports for mocks in `index.ts`
- do not exclude "test" folder in `tsconfig`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,9 +39,9 @@ knora-stack: ## runs the knora-stack
 generate-test-data: ## downloads generated test data from Knora-API
 	@rm -rf $(CURRENT_DIR)/.tmp/typescript
 	mkdir -p $(CURRENT_DIR)/.tmp/typescript
-	sleep 45
+	sleep 90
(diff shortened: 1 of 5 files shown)


tobiasschweizer:
@subotic I am going to figure out about the name of the dev package.

kilchenmann:
> @subotic I am going to figure out about the name of the dev package.

@tobiasschweizer When we follow semantic versioning, we should use a prereleas-tag: https://docs.npmjs.com/misc/semver#prerelease-tags and using the `npm version` command with `premajor` or `prerelease`. Read more heree: https://docs.npmjs.com/cli/version

tobiasschweizer:
After discussion with @kilchenmann:

We agreed that release candidates (with prerelease tag `rc.X`) will contain the mocks needed for development. The released productive version will be equivalent to the latest release candidate for a given version, but without the mocks.

PR #186, opened 2020-05-19
Generate test data

SepidehAlassi:
closes [DSP-229](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-29?issue=DSP-229)

- [x] create test data for ontology operations


SepidehAlassi:
@tobiasschweizer I cannot find why the tests fail. Tests also fail on [Andre's PR](https://github.com/dasch-swiss/knora-api-js-lib/pull/189). Is it a CI problem?

tobiasschweizer:
> @tobiasschweizer I cannot find why the tests fail. Tests also fail on [Andre's PR](https://github.com/dasch-swiss/knora-api-js-lib/pull/189). Is it a CI problem?

I checked the output on github-ci and it says:

> npm run integrate-v2-test-data /home/runner/work/knora-api-js-lib/knora-api-js-lib/.tmp/typescript/test-data

> @knora/api@1.0.0-rc.1 integrate-v2-test-data /home/runner/work/knora-api-js-lib/knora-api-js-lib

> node scripts/integrate-files-v2.js "/home/runner/work/knora-api-js-lib/knora-api-js-lib/.tmp/typescript/test-data"
> Error: ENOENT: no such file or directory, stat '/home/runner/work/knora-api-js-lib/knora-api-js-lib/.tmp/typescript/test-data/v2/ontologies/add-cardinalities-to-class-nothing.json'

https://github.com/dasch-swiss/knora-api-js-lib/pull/186/checks?check_run_id=709665333#step:8:1110

I think it does not find the file in the generated test data when trying to copy it.

I think you have to run the test against a tagged version of knora-api that is defined in the makefile. This tagged version must contain the changes needed to get the new test data. Preferably this would be a knora-api rc. 

From this PR I would expect the following changes:
- new entries in `scripts/v2-test-data-config.json`
- new test data files (those that were added in the config: v2 files have a compacted and a fully expanded version in JSON-LD)

I don't understand why admin test data files have changed.
Have you initialised the test data using `make init-db-test` before generating the test data?

This PR should not modify existing entries in `scripts/v2-test-data-config.json` because this will break the specs.

There can surely be a discussion where some files have to go (resources or search etc.) but this should be discussed in a different context. The reason why this seems unclear is because there is a utility for parsing responses that is used both for the resources and search route.

SepidehAlassi:
> I don't understand why admin test data files have changed.

the new dump of client API test data contained modified admin test data. I have just updated the test data on api-js-lib with the new dump. Because I believe the js-lib should always use the latest test data version.

> There can surely be a discussion where some files have to go (resources or search etc.) but this should be discussed in a different context. The reason why this seems unclear is because there is a utility for parsing responses that is used both for the resources and search route.

the changes I have made in the destination of couple of files in `scripts/v2-test-data-config.json` are just correcting obvious copy-paste mistakes that were in the script. I believe when a file is created in a specific folder of test-data, it should reside in the same folder in api-js-lib

SepidehAlassi:
@tobiasschweizer I have updated the paths to test data in spec files.

SepidehAlassi:
> This PR should not modify existing entries in `scripts/v2-test-data-config.json` because this will break the specs.

I have corrected the paths to test data in spec files.

tobiasschweizer:
> the new dump of client API test data contained modified admin test data. I have just updated the test data on api-js-lib with the new dump. Because I believe the js-lib should always use the latest test data version.

I have made a JS lib release yesterday with knora-api v13.0.0-rc.3 and could not see any changes.

> the changes I have made in the destination of couple of files in scripts/v2-test-data-config.json are just correcting obvious copy-paste mistakes that were in the script. I believe when a file is created in a specific folder of test-data, it should reside in the same folder in api-js-lib

This is not how it works. The config file maps between knora-api and the JS Lib.

SepidehAlassi:
> This is not how it works. The config file maps between knora-api and the JS Lib.
Please dump the data from the `clientapitest` and look for `things.json` and `things-with-paging`, they are clearly in `search` folder not in the resource. Hence the config file was mapping them incorrectly to the `resource` folder. That is what I have corrected. 

Furthermore, the new dump contains more test data for values that previously did not exist: `search/things-Link.json`, `resources/erase-resource-request.json`, `resources/thing-with-picture.json`, etc.

> I have made a JS lib release yesterday with knora-api v13.0.0-rc.3 and could not see any changes
Try dumping the test-data using develop branch, you will see the difference.

SepidehAlassi:
The purpose of this PR is not only generating ontology test data, but also in general updating client API test data as stated in the [PR description](https://github.com/dasch-swiss/knora-api-js-lib/pull/186#issue-420280032)

subotic:
> This PR should add new ontology test data. It 's not understandable to me how this affects admin test data.

I think there was a bug in the Admin-API that @SepidehAlassi corrected, which now leads to different data.

SepidehAlassi:
@kilchenmann  I have reversed my corrections of paths in the mapping script. That means the test data in `v2/resources` and `v2/search` folders in JS Lib are not identical to the ones generated from the knora-api. Mapping script incorrectly maps the test data generated from API to different paths in JS lib. Correction of mapping script is not sufficient as this problem roots deeper into the Spec files. I would suggest keeping in mind that in another PR, mapping script must be corrected and corresponding Spec files must be sanitized. 

I have also reversed the update of admin data as I was asked to. Even though the admin test data in JS lib are obviously out of date, but someone else should update them in another PR. 

This PR now only adds the test data for ontology operations. If you need further test data, please let me know I will generate them for you from the backend.

PR #191, opened 2020-05-28
Rename package

kilchenmann:
This PR renames the package from @knora/api to @dasch-swiss/dsp-js
Resolves [DSP-224](https://dasch.myjetbrains.com/youtrack/issue/DSP-224)


tobiasschweizer:
@kilchenmann Thanks for this PR! 

Could we try setting up an automated publication process before doing the renaming? Just in case we screw up a few times, this would still go in the old @knora/api.

test-framework/yalc.lock
```diff
@@ -4,6 +4,10 @@
     "@knora/api": {
```
tobiasschweizer, line 4: This is a left over that can be removed.

test-framework/package-lock.json
```diff
@@ -0,0 +1,13633 @@
+{
```
tobiasschweizer, line 1: I did not add `package-lock.json` of the test app to git because it could lead to problems when some dependency version change in the lib and is not included here. Also the test app should give us an idea if the lib integration works with a new Angular project thats created from scratch.

kilchenmann:
> Could we try setting up an automated publication process before doing the renaming? Just in case we screw up a few times, this would still go in the old @knora/api.

Yes this is planned ([DSP-302](https://dasch.myjetbrains.com/youtrack/issue/DSP-302)). I thought to do it after this PR to not screw up the old npm package @knora/api.

test-framework/package-lock.json
kilchenmann: Oh, I didn't saw that. I will delete it

tobiasschweizer:
> I thought to do it after this PR to not screw up the old npm package @knora/api.

:-) So we are going to screw up anyway ... the question is just where it hurts the least ;-)

kilchenmann:
> the question is just where it hurts the least ;-)

I thought the new one. Because it doesn't exist yet. Nobody knows it. So if there's a mistake, no one will know about it. And we can [reverse the publication](https://docs.npmjs.com/unpublishing-packages-from-the-registry).

tobiasschweizer:
> I thought the new one. Because it doesn't exist yet. Nobody knows it. So if there's a mistake, no one will know about it. And we can [reverse the publication](https://docs.npmjs.com/unpublishing-packages-from-the-registry).

Sounds fair!

However, we couldn't use the same version again: 
> Once a package is unpublished, it cannot be republished. If you’ve unpublished a package by mistake, we’d recommend publishing again under a different name, or for unpublished versions, bumping the version number and publishing again.

https://docs.npmjs.com/unpublishing-packages-from-the-registry#when-to-unpublish

So I guess we could just start with 0.0.1-rc.0 or something like this.

tobiasschweizer:
I will briefly test this PR locally.

kilchenmann:
> So I guess we could just start with 0.0.1-rc.0 or something like this.

Yes, good point.

kilchenmann:
> Could we work on the automated publication next week or so? We still need to figure out what to do about the dev package including test data that we need for development of UI lib but don't need in production.

Yes, I thought I needed your help anyway. Especially with the dev package. It's planned for Sprint-6.

tobiasschweizer:
> Yes, I thought I needed your help anyway. Especially with the dev package. It's planned for Sprint-6.

And I need your help. Sounds like a good match :-)

PR #194, opened 2020-06-04
Fix bug ApiResonseError

tobiasschweizer:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-345

### DIFF
--- a/src/models/api-response-error.ts
+++ b/src/models/api-response-error.ts
@@ -53,8 +53,10 @@ export class ApiResponseError extends ApiResponse {
 
         const response = new ApiResponseError();
 
-        if (ajaxError.request.method) response.method = ajaxError.request.method;
-        if (ajaxError.request.url) response.url = ajaxError.request.url;
+        if (ajaxError.request) {
+            if (ajaxError.request.method) response.method = ajaxError.request.method;
+            if (ajaxError.request.url) response.url = ajaxError.request.url;
+        }
         if (ajaxError.xhr) response.status = ajaxError.xhr.status;
 
         response.error = ajaxError;
--- a/src/models/api-response-error.ts
+++ b/src/models/api-response-error.ts
@@ -49,7 +49,12 @@ export class ApiResponseError extends ApiResponse {
     /**
      * Create an instance from an AjaxError.
      */
-    static fromAjaxError(ajaxError: AjaxError): ApiResponseError {
+    static fromAjaxError(ajaxError: AjaxError | ApiResponseError): ApiResponseError {
+
+        // check if the error is already a ApiResponseError
+        if (ajaxError instanceof ApiResponseError) {
+            return ajaxError;
+        }
 
         const response = new ApiResponseError();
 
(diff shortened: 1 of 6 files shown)


src/models/api-response-error.ts
```diff
@@ -49,7 +49,12 @@ export class ApiResponseError extends ApiResponse {
     /**
      * Create an instance from an AjaxError.
      */
-    static fromAjaxError(ajaxError: AjaxError): ApiResponseError {
+    static fromAjaxError(ajaxError: AjaxError | ApiResponseError): ApiResponseError {
+
+        // check if the error is already a ApiResponseError
+        if (ajaxError instanceof ApiResponseError) {
```
tobiasschweizer, line 55: @andreas-aeschlimann This took me a while to figure out. 

@loicjaouen experienced a problem where `ApiResponseError` could not be instantiated correctly.

I now found out that an `ApiResponseError` can be instantiated from an existing `ApiResponseError`.
An example would be the resources-endpoint:

https://github.com/dasch-swiss/knora-api-js-lib/blob/7c38e545ec026e1707482152668a7b3e76183f83/src/api/v2/resource/resources-endpoint-v2.ts#L34-L72

When `getResource()` is called, it internally calls `getResources`. If the HTTP response is not successful, then `handleError` is called **twice** because there are two `catchError` blocks.

tobiasschweizer:
@loicjaouen I think I found the root of the problem.

I will add some more unit tests for failed HTTP requests in different endpoints and then this PR should be ready for review.

I will give you instructions how you can test it locally (local build).

tobiasschweizer:
@loicjaouen Could you please try this PR in your local setup?

Steps to build this locally:
- clone this repo
- checkout wip/dsp-345-error-handling-http
- from the project root, run:
    - `npm i`
    - `npm run peer-deps`
    - `npm run yalc-publish` (make sure `yalc` is installed: `npm install yalc -g`)
- in your project (Angular app etc):
    - `yalc remove --all && yalc add @dasch-swiss/dsp-js`

Note that the lib is already called `@dasch-swiss/dsp-js` although not published yet. We are in the middle of renaming and a first publication on npmjs.com under the new name will happen asap.

loicjaouen:
I also needed to (https://github.com/loicjaouen/GettingRepresentations/commit/a8b7d4fe5c79dae2c86ec9fc3097c690724fa0c8):
- remove the packaged installed by npm: `npm remove @knora/api@1.0.0-rc.1`
- delete it from `package.json` (I thought the above would do that)
- and change in the code to import `from "@dasch-swiss/dsp-js";`

and running the offending code:
- `ERROR TypeError: "ajaxError.request is undefined"` is not showing up anymore
- error is caught and can be managed by client's code ( https://github.com/loicjaouen/GettingRepresentations/blob/master/src/app/app.component.ts#L42 )

Thanks :)

tobiasschweizer:
> this code does what is expected from it

I'd say that's a compliment :blush:

PR #196, opened 2020-06-09
ontology IRI: ignore port if not running on localhost

loicjaouen:
follow knora's recipe for forging an IRI: https://github.com/dasch-swiss/knora-api/blob/0127dcae27e46bc3a4c9947282c30038c0a25d20/webapi/src/main/scala/org/knora/webapi/Settings.scala#L54

### DIFF
--- a/src/models/v2/ontologies/OntologyConversionUtil.spec.ts
+++ b/src/models/v2/ontologies/OntologyConversionUtil.spec.ts
@@ -11,64 +11,74 @@ describe("OntologyConversionUtil", () => {
         jasmine.Ajax.uninstall();
     });
 
-    describe("Method getOntologyIriFromEntityIri", () => {
-        const localConfig = new KnoraApiConfig("http", "0.0.0.0", 3333);
-        const remoteConfig = new KnoraApiConfig("https", "test.example.com", 3333);
+    describe("Method getOntologyIriFromEntityIri with local setup", () => {
+        const config = new KnoraApiConfig("http", "0.0.0.0", 3333);
 
         it("should get the ontology IRI from a knora-api entity", () => {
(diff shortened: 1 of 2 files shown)


tobiasschweizer:
@loicjaouen I would be great if you could add a test case to `src/models/v2/ontologies/OntologyConversionUtil.spec.ts`.

You can copy the existing `describe`, add it below and adapt the `apiHost` setting for `KnoraApiConfig`. 

Thanks a lot!

loicjaouen:
apologies @tobiasschweizer I only now read your comment and I added the tests differently, do you mind?

tobiasschweizer:
> I only now read your comment and I added the tests differently, do you mind?

Could you make it two `describe` blocks, each having their own config object? And then they just contain the same `it`s, with different expectations of course.

```ts
describe("Method getOntologyIriFromEntityIri with local setup", () => {

        const config = new KnoraApiConfig("http", "0.0.0.0", 3333);

        it("should get the ontology IRI from a knora-api entity", () => {
          ...
        });

}

describe("Method getOntologyIriFromEntityIri with remote setup", () => {

        const config = new KnoraApiConfig("https", "test.example.com", 3333);

        it("should get the ontology IRI from a knora-api entity", () => {
          ...
        });

}
```

This way it's very clear that in one case your are testing a local config and a remote config in the other.

Sorry for being annoying with this ...

loicjaouen:
no problem, as you wish :)

tobiasschweizer:
Ok, good to go! Thanks a lot, @loicjaouen.

loicjaouen:
thank you @tobiasschweizer and @subotic  :)

tobiasschweizer:
@loicjaouen I merged your PR, I hope you don't mind.

loicjaouen:
perfect, thanks (I updated the branch and was waiting for the test to complete)

PR #202, opened 2020-06-18
feature (ontology endpoint): create ontology

tobiasschweizer:
Example for a method on ontology endpoint.

### DIFF
--- a/src/api/v2/ontology/ontologies-endpoint.spec.ts
+++ b/src/api/v2/ontology/ontologies-endpoint.spec.ts
@@ -200,7 +200,7 @@ describe("OntologiesEndpoint", () => {
 
     describe("Method createOntology", () => {
 
-        it("should create a new ontology", () => {
+        it("should create a new ontology", done => {
 
             const newOnto = new CreateOntology();
             newOnto.label = "The foo ontology";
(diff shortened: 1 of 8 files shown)


src/api/v2/ontology/ontologies-endpoint.spec.ts
```diff
@@ -196,4 +198,36 @@ describe("OntologiesEndpoint", () => {
...
+            newOnto.projectIri = "http://rdfh.ch/projects/00FF";
+
+            knoraApiConnection.v2.onto.createOntology(newOnto).subscribe((ontoCreationRes: CreateOntologyResponse) => {
+                expect(ontoCreationRes.label).toEqual("The foo ontology");
+                done();
+            });
+
+            const request = jasmine.Ajax.requests.mostRecent();
+
+            const ontoCreationResponse = require("../../../../test/data/api/v2/ontologies/create-empty-foo-ontology-expanded.json");
```
tobiasschweizer, line 217: @kilchenmann I don't know whether this is the mocked response to an ontology creation request. The last mod date is missing and it looks like the request. Is this the wrong file?

tobiasschweizer: currently the unit tests are failing because of that:

ERROR: Error: Fatal error in JsonConvert. Failed to map the JSON object to the class "CreateOntologyResponse" because the defined JSON property "http://api.knora.org/ontology/knora-api/v2#lastModificationDate" does not exist:

	Class property: 
		lastModificationDate

	JSON property: 
		http://api.knora.org/ontology/knora-api/v2#lastModificationDate

tobiasschweizer: and there seems to be another problem with the copying of the test data files

PR #208, opened 2020-06-27
Create and delete ontology

kilchenmann:
Resolves [426](https://dasch.myjetbrains.com/youtrack/issue/DSP-426)

This PR uses Knora/DSP-API v13.0.0-rc.8.
Create and delete an ontology is the main part of this PR.

You can test it in test-framework:
1. Get the latest version of the branch: `git pull`
1. Install all depndencies: `npm install && npm run peer-deps`
1. Prepare and build the lib: `npm run prepare-dev-publication && npm run yalc-publish`
    (prepare-dev-publication will update 3 files. `git status` will show them as changed files. Please do not commit or push them!)
1. Go to test-framework: `cd test-framework`
1. Prepare test-framework (from dsp-js-lib test-framework folder): `yalc remove --all && yalc add @dasch-swiss/dsp-js && npm install`
1. Run the test-framework app: `ng s`
1. Open browser and go to: <http://localhost:4200>
1. Be sure that Knora/DSP-API is running in the background
1. Click on `LOGIN` button
1. Go to section "Ontology" and click the buttons in the following order:
  * `Create testonto` will create a test ontology in anything project
  * `Get testonto` will grab the data of new created test ontology
  * Optional: Click on `Get anything project ontologies` in section "Project specific ontologies" and you should get 4 instead of 3 ontologies
  * `Delete testonto` will delete the testonto
  * Optional: Click on `Get anything project ontologies` in section "Project specific ontologies" and you should get 3 ontologies again

PR #211, opened 2020-07-02
Use knora-api 13.0.0-rc.8

kilchenmann:
Use [knora-api 13.0.0-rc.8](https://github.com/dasch-swiss/knora-api/releases/tag/v13.0.0-rc.8) with Fuseki instead GraphDB.


kilchenmann:
There's a known bug in knora-api, which doesn't allow to generate testdata with `clientapitest`. This PR has to wait...

subotic:
@kilchenmann Bug is fixed in knora-api 13.0.0-rc.7

kilchenmann:
Probably I will use this PR to upgrade to rc.8 directly which is needed for PR #208.

kilchenmann:
I need the changes from this PR for PR #208.
Can you please review it as soon as possible @subotic or @tobiasschweizer? Thanks

tobiasschweizer:
Ok, I will review this PR this morning.

Makefile
```diff
@@ -29,7 +29,7 @@ endef
 # Clones the knora-api git repository
 .PHONY: clone-knora-stack
 clone-knora-stack:
-	@git clone --branch v13.0.0-rc.5 --single-branch --depth 1 https://github.com/dasch-swiss/knora-api.git $(CURRENT_DIR)/.tmp/knora-stack
+	@git clone --branch v13.0.0-rc.8 --single-branch --depth 1 https://github.com/dasch-swiss/knora-api.git $(CURRENT_DIR)/.tmp/knora-stack
```
tobiasschweizer, line 32: Are you sure that the commands to start the stack without the api, reload the data, and then start the api are still correct?

It seems to recompile the code when reloading the data with `init-db-test` although it should have already compiled in a previous step.

test/data/api/admin/groups/get-group-members-response.json
```diff
@@ -1 +1 @@
...
...tion sowie für die Erforschung der Bilder und Texte nutzbar machen. Auch können bereits bestehende und entstehende Online-Editionen damit verknüpft  werden , wodurch die Nutzung von Datenbanken anderer Institutionen im Hinblick auf unser Corpus optimiert wird.\n</p>"}],"id":"http://rdfh.ch/projects/0803","keywords":["Basel","Basler Frühdrucke","Bilderfolgen","Contectualisation of images","Inkunabel","Kunsthistorisches Seminar Universität Basel","Late Middle Ages","Letterpress Printing","Narrenschiff","Sebastian Brant","Wiegendrucke","early print","incunabula","ship of fools"],"logo":"incunabula_logo.png","longname":"Bilderfolgen Basler Frühdrucke","ontologies":["http://www.knora.org/ontology/0803/incunabula"],"selfjoin":false,"shortcode":"0803","shortname":"incunabula","status":true},{"description":[{"value":"A demo project of a collection of images","language":"en"}],"id":"http://rdfh.ch/projects/00FF","keywords":["collection","images"],"logo":null,"longname":"Image Collection Demo","ontologies":["http://www.knora.org/ontology/00FF/images"],"selfjoin":false,"shortcode":"00FF","shortname":"images","status":true}],"sessionId":null,"status":true,"token":null,"username":"multiuser"}]}
```
tobiasschweizer, line 1: I think the reason for the difference is that order of projects is different: demo project and incunabula.

kilchenmann: Yes, but it doesn't matter? Or will it always be different?

tobiasschweizer: I think as long as the order is not defined it does not matter. And the format is still the same. I just wondered why it would be different.

Makefile
kilchenmann: You mean we do not have to run `make stack-restart-api` after `make init-db-test`? 
@subotic Do we need a different setup with fuseki now?

kilchenmann: This is `knora-stack` at the moment:

```make
.PHONY: knora-stack
knora-stack: ## runs the knora-stack
	$(MAKE) -C $(CURRENT_DIR)/.tmp/knora-stack stack-without-api
	$(MAKE) -C $(CURRENT_DIR)/.tmp/knora-stack print-env-file
	$(MAKE) -C $(CURRENT_DIR)/.tmp/knora-stack stack-config
	sleep 15
	$(MAKE) -C $(CURRENT_DIR)/.tmp/knora-stack init-db-test
	sleep 15
	$(MAKE) -C $(CURRENT_DIR)/.tmp/knora-stack stack-restart-api
	sleep 35
	$(MAKE) -C $(CURRENT_DIR)/.tmp/knora-stack stack-logs-api-no-follow
```

tobiasschweizer: I don't know, but the commands seem to behave differently now.

kilchenmann: Resolved in 36f9d1c

Issue #122, opened 2019-12-10
check dockerfile for testframework


tobiasschweizer:
see path in https://github.com/dasch-swiss/knora-api-js-lib/pull/119#pullrequestreview-329020891

PR #215, opened 2020-07-09
Mock resources

kilchenmann:
This PR contains:

- refactor file names for mocked test data
- export of additional mocked test data: list of resources as ReadResourceSequence; needed for [PR #122 in ui-lib](https://github.com/dasch-swiss/dsp-ui-lib/pull/122)


tobiasschweizer:
@kilchenmann What is this PR about?

tobiasschweizer:
Would you like to resolve https://github.com/dasch-swiss/knora-api-js-lib/issues/62?

kilchenmann:
> Would you like to resolve #62?

Not primary. I wanted to add a resources-mock file in test/data and saw the different names for the other mock-files. So, the first two commits here were only to make the file names more consistent.

> What is this PR about?

I have to mock a ReadResourceSequence to use for the tests in ui-lib. I will add a description to the PR soon. 
Let me know if this is the wrong way. We talked about it in the scrum meeting today.

tobiasschweizer:
> > Would you like to resolve #62?
> 
> Not primary. I wanted to add a resources-mock file in test/data and saw the different names for the other mock-files. So, the first two commits here were only to make the mock data file names more consistent.
> 
> > What is this PR about?
> 
> I have to mock a ReadResourceSequence to use for the tests in ui-lib. I will add a description to the PR soon.
> Let me know if this is the wrong way. We talked about it in the scrum meeting today.

I understand. Sorry, I didn't get it in the first place. Maybe you could just make a mock that returns a `ReadResourceSequence` and internally calls `MockResourcegetTestthing()`. Just make sure that the sequence contains **different** instances of testthing (objects are reference types). Just call `MockResource.getTestthing()` as many times as there are elements in the mocked `ReadResourceSequence `.

Otherwise I can help you with that next week.

test/data/api/v2/mock-resources.ts
```diff
@@ -0,0 +1,61 @@
...
+import { ListNodeV2Cache } from "../../../../src/cache/ListNodeV2Cache";
+import { OntologyCache } from "../../../../src/cache/ontology-cache/OntologyCache";
+import { KnoraApiConfig } from "../../../../src/knora-api-config";
+import { ReadResourceSequence } from "../../../../src/models/v2/resources/read/read-resource-sequence";
+import { ResourcesConversionUtil } from "../../../../src/models/v2/resources/ResourcesConversionUtil";
+import testthings from "../v2/resources/things-expanded.json";
+import { MockList } from "./mock-list";
+import { MockOntology } from "./mock-ontology";
+
+export namespace MockResources {
```
tobiasschweizer, line 15: I think you could just add a new method to the existing `MockResource`:

```ts
    export const getTesthings = (length = 25): Observable<ReadResourceSequence> => {

        const resources: Array<Observable<ReadResource>> = new Array(length).fill(0).map(
            () => {
                return getTestthing();
            }
        );

        return forkJoin(resources).pipe(
            map(
                (res: ReadResource[]) => {
                    return new ReadResourceSequence(res);
                }
            )
        );

    };
```

You could even add an arg to set `mayHaveMoreResults` on `ReadResourceSequence` to make it more flexible.

tobiasschweizer: this is some `RxJS` magic: https://www.learnrxjs.io/learn-rxjs/operators/combination/forkjoin

kilchenmann: Resolved in 723dfbb and 63d2a25

test/data/api/v2/mock-resource.ts
```diff
@@ -59,4 +59,22 @@ export namespace MockResource {
 
     };
 
+    export const getTesthings = (length = 25): Observable<ReadResourceSequence> => {
```
tobiasschweizer, line 62: If you want, you could pass a Boolean arg that would set `mayHaveMoreResults`. Maybe that could be useful for testing the search result component?

test/data/api/v2/mock-resources.ts
kilchenmann: Nice solution btw

test/data/api/v2/mock-resource.ts
kilchenmann: Yes, why not?

test/data/api/v2/mock-resource.ts
```diff
@@ -59,7 +59,7 @@ export namespace MockResource {
 
     };
 
-    export const getTesthings = (length = 25): Observable<ReadResourceSequence> => {
+    export const getTesthings = (length = 25, mayHaveMoreResults = false): Observable<ReadResourceSequence> => {
```
kilchenmann, line 62: @tobiasschweizer That should work like this, right?

test/data/api/v2/mock-resource.ts
kilchenmann: Done in 
af0d0d6

test/data/api/v2/mock-resource.ts
```diff
@@ -59,4 +59,22 @@ export namespace MockResource {
...
+        const resources: Array<Observable<ReadResource>> = new Array(length).fill(0).map(
+            () => {
+                return getTestthing();
+            }
+        );
+
+        return forkJoin(resources).pipe(
+            map(
+                (res: ReadResource[]) => {
+                    return new ReadResourceSequence(res, mayHaveMoreResults);
```
kilchenmann, line 73: @tobiasschweizer That should work like this, right?

tobiasschweizer: Yes, I think so.

kilchenmann:
> I think this PR is good to go now.

Thanks

PR #223, opened 2020-08-05
Update process of release notes

kilchenmann:
This PR splits the Gren workflow step into two variations: one runs on prerelease with the `--prerelease` option and one on release without this parameter.
The options parameter is a new feature of the latest release of [lakto/gren-action](https://github.com/lakto/gren-action/tree/v2.0.0)


kilchenmann:
It's the same as in [ui-lib](https://github.com/dasch-swiss/dsp-ui-lib/pull/145)

PR #226, opened 2020-08-06
Prerelease v1.0.0 rc.6

kilchenmann:


kilchenmann:
@flavens This time I updated the `CHANGELOG.md` with GREN. This is why it has so many changes. We can use `gren changelog` instead of `gren release` to update the changelog file.

kilchenmann:
@mdelez Thanks

PR #232, opened 2020-08-20
Add missing declarations

flavens:
closes [DSP-559](https://dasch.myjetbrains.com/youtrack/issue/DSP-559)

### DIFF
--- a/src/models/v2/resources/ResourcesConversionUtil.ts
+++ b/src/models/v2/resources/ResourcesConversionUtil.ts
@@ -374,7 +374,11 @@ export namespace ResourcesConversionUtil {
                 case Constants.LinkValue: {
                     const linkVal = handleLinkValue(valueJsonld, ontologyCache, listNodeCache, jsonConvert);
                     value = linkVal.pipe(map((val: ReadLinkValue) => {
-                        val.strval = val.linkedResourceIri;
+                        if (val.linkedResource !== undefined) {
+                            val.strval = val.linkedResource.label;
+                        } else {
+                            val.strval = val.linkedResourceIri;
+                        }
                         return val;
                     }));
                     break;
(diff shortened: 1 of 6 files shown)


src/models/v2/resources/read/read-resource.ts
```diff
@@ -52,6 +52,12 @@ export class ReadResource extends ReadWriteResource {
 
     outgoingReferences: ReadResource[] = [];
 
+    incomingRegions: ReadResource[] = [];
+
+    incomingStillImageRepresentations: ReadResource[] = [];
+
+    stillImageRepresentationsToDisplay: ReadResource[] = [];
```
tobiasschweizer, line 59: This should be `StillImageRepresentation[]` but it can be handled in BEOL app

tobiasschweizer: This can be handled in BEOL app until there is a generic solution for this.

src/models/v2/resources/ResourcesConversionUtil.ts
```diff
@@ -374,7 +374,11 @@ export namespace ResourcesConversionUtil {
                 case Constants.LinkValue: {
                     const linkVal = handleLinkValue(valueJsonld, ontologyCache, listNodeCache, jsonConvert);
                     value = linkVal.pipe(map((val: ReadLinkValue) => {
-                        val.strval = val.linkedResourceIri;
+                        if (val.linkedResource !== undefined) {
+                            val.strval = val.linkedResource.label;
```
tobiasschweizer, line 378: @lrosenth if the referred resource is contained in the link value, the `strval` will contain its label, otherwise it will be set to its Iri.

mdelez (approved):
LGTM

PR #235, opened 2020-09-09
DSP-428 Create res properties

kilchenmann:
Resolves [DSP-428](https://dasch.myjetbrains.com/youtrack/issue/DSP-428)


src/models/v2/ontologies/create/create-resource-property.ts
```diff
@@ -0,0 +1,36 @@
+import { JsonObject, JsonProperty } from "json2typescript";
+import { StringLiteral } from "../../../admin/string-literal";
+import { IdConverter } from "../../custom-converters/id-converter";
+import { UpdateOntology } from "../update-ontology";
+
+@JsonObject("CreateResourceProperty")
```
tobiasschweizer, line 6: @kilchenmann If I understand correctly, only the classes with he suffix "payload" are actually going to be sent to knora-api. So `CreateResourcePropertyPayload` actually has to be serialised to JSON-LD, whereas `CreateResourceProperty` is expected as an argument by the method `createResourceProperty` and an instance of `CreateResourcePropertyPayload` is then created from it.

Why would we need to have TS decorators for `json2typescript` on `CreateResourceProperty` then? Sicne this is not going to be serialized directly, we could even add a constructor to make it easier for the user to create an instance of `CreateResourceProperty `.

kilchenmann: Yes of course, this would be the better way. Thanks.

src/api/v2/ontology/ontologies-endpoint-v2.ts
```diff
@@ -151,7 +161,7 @@ export class OntologiesEndpointV2 extends Endpoint {
         const newResClass = new NewResourceClass();
         newResClass.id = resClass.ontology.id + Constants.Delimiter + resClass.name;
         newResClass.label = resClass.labels;
-        newResClass.comment = resClass.comments;
+        newResClass.comment = (resClass.comments.length ? resClass.comments : resClass.labels);
```
tobiasschweizer, line 164: @kilchenmann Is this supposed to be a fallback if no comments are provided? But then how would we know that the labels array is not empty?

tobiasschweizer:
@kilchenmann tomorrow I will complete the test for deleting an ontology.

The DELETE request does not have a body, and the response will be the ontology's metadata (https://docs.knora.org/03-apis/api-v2/ontology-information/#deleting-a-property).

tobiasschweizer:
@kilchenmann I think the test for property deletion is fine: https://github.com/dasch-swiss/dsp-js-lib/pull/235/commits/2cdeb527474756e21e41987652217bd97f658465

tobiasschweizer:
@kilchenmann I will add the missing e2e tests and then this PR is ready for the final review.

kilchenmann:
Sounds great, thanks!

tobiasschweizer:
@kilchenmann I created https://dasch.myjetbrains.com/youtrack/issue/DSP-685. I will work on this once this PR is merged.

Could please check if this PR implements what is required for https://dasch.myjetbrains.com/youtrack/issue/DSP-428? You won't be to review this PR I think since you initially created it.

tobiasschweizer (approved):
I think I did a great job.

src/api/v2/ontology/ontologies-endpoint-v2.ts
kilchenmann: Yes, this is a kind of fallback. The labels are a mandatory value, they shouldn't be empty. But I think we do not check that, do we?

src/api/v2/ontology/ontologies-endpoint-v2.ts
```diff
@@ -196,4 +204,99 @@ export class OntologiesEndpointV2 extends Endpoint {
...
+
+        newResProperty.id = resProp.ontology.id + Constants.Delimiter + resProp.name;
+
+        newResProperty.label = resProp.labels;
+        newResProperty.comment = (resProp.comments.length ? resProp.comments : resProp.labels);
+        newResProperty.subPropertyOf = resProp.subPropertyOf;
+        newResProperty.type = Constants.ObjectProperty;
+
+        newResProperty.subjectType = resProp.subjectType;
+        newResProperty.objectType = resProp.objectType;
```
kilchenmann, line 231: Can we simplify the `CreateResourceProperty` or at least make the `objectType` optional. If `objectType` is not defined we can define it from `subPropertyOf`.

```typescript
// If the new property is a subproperty of knora-api:hasValue, its knora-api:objectType must be one of the built-in subclasses of knora-api:Value.
// If the new property is a subproperty of knora-base:hasLinkTo, its knora-api:objectType must be a subclass of knora-api:Resource

newResProperty.objectType = (resProp.subPropertyOf === Constants.HasValue ? Constants.Value : Constants.Resource)
```

Not sure if all Constants are defined. And maybe this is not a good solution as the `subPropertyOf`could be a subproperty of the mentioned `knora-api:hasValue` and `knora-api:hasLinkTo`.
Or what do you think @tobiasschweizer?

src/models/v2/ontologies/create/create-resource-property.ts
tobiasschweizer: I will do this as part of https://dasch.myjetbrains.com/youtrack/issue/DSP-685

src/api/v2/ontology/ontologies-endpoint-v2.ts
tobiasschweizer: The problem is that with `json2typescript`, all properties have an initial value, so technically speaking, all props are set. But the initial value does not make any sense, e..g, an empty string.

So yes, there could be some checks in the endpoint's method for empty strings and empty arrays.

src/api/v2/ontology/ontologies-endpoint-v2.ts
tobiasschweizer: I think we could do some consistency checks for value and link props. Or we could make specific classes for value and link props. Would that help?

src/api/v2/ontology/ontologies-endpoint-v2.ts
kilchenmann: Should we implement additional checks in this PR? I suggest to do it in a separate one..

src/api/v2/ontology/ontologies-endpoint-v2.ts
kilchenmann: A user of the method needs a deep knowledge of the ontology stuff. I thought that the object you hand over to the method could be simplified. At this point it is similar to the payload. And you have to take care that you pass the correct `object-type` matching the `subPropertyOf`. 
I had this idea already in mind when I created the method and the `CreateResourceProperty` class and thought to clean it up at the end.

I'll be able to use it in DSP-App. But if somebody else uses it, it gets complicated very fast. For the moment it's ok for me. It was just an idea / suggestion.

src/api/v2/ontology/ontologies-endpoint-v2.ts
tobiasschweizer: I will do it in https://dasch.myjetbrains.com/youtrack/issue/DSP-685, I added a note

src/api/v2/ontology/ontologies-endpoint-v2.ts
tobiasschweizer: It's definitely worth thinking about. I will keep that in kind when doing the refactoring in https://dasch.myjetbrains.com/youtrack/issue/DSP-685

tobiasschweizer: I made a note in https://dasch.myjetbrains.com/youtrack/issue/DSP-685

tobiasschweizer:
@kilchenmann Do you think this PR can be merged so I can start the refactoring?

kilchenmann:
> @kilchenmann Do you think this PR can be merged so I can start the refactoring?

Yes I think so. Let's merge it

PR #253, opened 2020-11-05
Read an Islamic Date

tobiasschweizer:
resolves DSP-941

This PR makes the era (both start and and) optional when deserialising from and serialising to JSON-LD:
- reading a date
- creating a date
- updating a date

The era remains a required property in `KnoraDate`. The absence of an era is indicated by "noEra".

This is because if an era is missing from a Christian date, CE / AD can be assumed. But this should be dealt with in dsp-js, not further down. To make this consistent, "noEra" is set if the given calendar date does not support the concept of an era. Making the era optional on `KnoraDate` would require the client to check for its existence when accessing it, to avoid errors when absent (undefined).

In a subsequent PR, the Islamic date can be activated in the calendar header in dsp-ui. Also era handling has to be adapted when creating / updating Islamic dates from dsp-ui. Currently, "CE" is hard-coded in `DateInputComponent`. Also `DateValueComponent.populateValue` has to be adapted. -> https://dasch.myjetbrains.com/youtrack/issue/DSP-1095

**Open Question**: Test date for the Islamic date was added manually, and in consequence, the build process had to be adapted in the Makefile. The automatic data generation step is skipped now. We should find a permanent solution here.

**EDIT**: I think the manually generated data should be kept in a separate folder since the test data will still be integrated from the assets (#257 ). -> done in https://github.com/dasch-swiss/dsp-js-lib/pull/253/commits/499b61b55abf1639160bf12677760a7916b04e97


subotic (approved):
Looks good, thanks!

PR #254, opened 2020-11-05
build: depend on main

subotic:
resolves DSP-


kilchenmann:
@subotic I think this PR is not needed for the moment. We could close it without merging

subotic:
yes, you are right.

PR #255, opened 2020-11-09
DSP-978 generate changelog automatically

kilchenmann:
resolves DSP-978


tobiasschweizer (approved):
So this is like the PR already done in dsp-ui?

kilchenmann:
Yes same as in dsp-ui and dsp-app

PR #256, opened 2020-11-11
refactor(ontology): label and comment should not be optional in resource class (DSP-927)

mdelez:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-927


mdelez:
The tests are failing because the test data for properties do not contain comments. Although the docs say that a comment is required so each property should have a comment (or at least an empty string).

https://docs.knora.org/03-apis/api-v2/ontology-information/#creating-a-property

PR #257, opened 2020-11-11
Patch Expected Test Data

tobiasschweizer:
Patch expected data so that it conforms to the expectation. This is a workaround to deal with Iris and timestamps in JSON test data that are differ from rc to rc.


mdelez (approved):
that commit message tho 👌

subotic (approved):
looks good.

PR #261, opened 2020-11-18
DSP-915 - Missing lastModificationDate property 

mdelez:
resolves DSP-915


mdelez:
@kilchenmann I added the lastModificationDate property to both classes. The timestamp is grabbed from the OntologyMetadata (which is updated any time a resource class or property is modified) so maybe the property should be renamed to "ontoLastModificationDate" to make it clear that the timestamp does not correlate to when a specific resource class or property was modified but instead the last time the ontology was modified? I think it would also be smart to have `lastModificationDate`s for each resource class and property but this would be a new task.

kilchenmann (approved):
It look good and it works. Thanks!

PR #266, opened 2020-12-04
DSP-1138 Add missing guiOrder to serialize IHasProperty

kilchenmann:
resolves DSP-1138

For the reviewers. You can have a look in the updated file in `deserialize` that there we have `guiOrder` definition but it was missing in `serialize`


mpro7 (approved):

tobiasschweizer (commented):
That looks good.

Is there some way we could test this? I think the [(unit) test data](test/data/api/v2/ontologies/add-cardinalities-to-class-nothing-request.json) does contain not a gui order.

tobiasschweizer (approved):

PR #269, opened 2020-12-09
DSP-883 Support New Edit List Endpoints

mdelez:
resolves DSP-883


tobiasschweizer:
Would it be ok to use the generated test data once #270 is merged?

mdelez:
> Would it be ok to use the generated test data once #270 is merged?

done in cef6e66

src/api/admin/lists/lists-endpoint.spec.ts
```diff
@@ -198,6 +202,121 @@ describe("ListsEndpoint", () => {
...
+
+            const listItemIri = "http://rdfh.ch/lists/0001/treeList01";
+
+            const newName = "updated third child name";
+
+            childNodeInfo.name = newName;
+
+            knoraApiConnection.admin.listsEndpoint.updateChildName(listItemIri, childNodeInfo).subscribe(
+                (res: ApiResponseData<ChildNodeInfoResponse>) => {
+                    done();
```
tobiasschweizer, line 219: Could you add a basic assertion like:

```ts
expect(res.body.nodeinfo.id).toEqual("http://rdfh.ch/lists/0001/treeList01");
```

I know that not all the `it`s of this endpoint's specs have it, so if you are up to it, you could add some more ;-)

src/api/admin/lists/lists-endpoint.spec.ts
```diff
@@ -198,6 +202,121 @@ describe("ListsEndpoint", () => {
...
+
+            const newLabels = new StringLiteral();
+            newLabels.language = "se";
+            newLabels.value = "nya märkningen för nod";
+
+            childNodeLabels.labels = [newLabels];
+
+            knoraApiConnection.admin.listsEndpoint.updateChildLabels(listItemIri, childNodeLabels).subscribe(
+                (res: ApiResponseData<ChildNodeInfoResponse>) => {
+                    done();
```
tobiasschweizer, line 258: same as above

src/api/admin/lists/lists-endpoint.spec.ts
```diff
@@ -198,6 +202,121 @@ describe("ListsEndpoint", () => {
...
+
+            const newComments = new StringLiteral();
+            newComments.language = "se";
+            newComments.value = "nya kommentarer för nod";
+
+            childNodeComments.comments = [newComments];
+
+            knoraApiConnection.admin.listsEndpoint.updateChildComments(listItemIri, childNodeComments).subscribe(
+                (res: ApiResponseData<ChildNodeInfoResponse>) => {
+                    done();
```
tobiasschweizer, line 297: same as above

tobiasschweizer (commented):
Looks good!

Do you think it would make sense to add calls to these new methods in the test app? Then there could also be E2E tests for these. The admin API does just have a few E2E tests so far.

In general, E2E tests should be split up by type, see https://dasch.myjetbrains.com/youtrack/issue/DSP-684. But this we have to do in a second step.

tobiasschweizer:
Is there support need for "`PUT: /admin/lists/<nodeIri>/position"?

mdelez:
> Is there support needed for "`PUT: /admin/lists//position"?

Yes, but not in this PR. I've changed the title of this PR to make it more clear.

src/api/admin/lists/lists-endpoint.spec.ts
mdelez: aren't the same things checked when doing this assertion? `expect(request.data()).toEqual(payload);`

mdelez:
@tobiasschweizer e2e tests have been added, please let me know if I did something incorrectly

tobiasschweizer (approved):
Great, thanks!

We should really work on splitting the E2E tests per type.

PR #286, opened 2021-01-06
feat(endpoint): Retry failed HTTP requests (DSP-170)

tobiasschweizer:
resolves DSP-170

This PR introduces a custom RxJS operator called `retryOnError` (inspired by https://medium.com/angular-in-depth/retry-failed-http-requests-in-angular-f5959d486294).

The operator retries failed HTTP request for specified error status codes. If the error is the user's fault (malformed query etc.), no retry should be performed. Retries should happen when the network is unstable ([XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) [status 0](https://stackoverflow.com/questions/47424413/rxjs-5-how-to-get-statuscode-on-catch)):

retry

### DIFF
--- a/src/api/operators/retryOnError.ts
+++ b/src/api/operators/retryOnError.ts
@@ -16,7 +16,8 @@ import { delay, mergeMap, retryWhen, tap } from "rxjs/operators";
 export function retryOnError(delayMs: number, maxRetries: number, retryOnErrorStatus: number[], logError: boolean) {
     let retries = maxRetries;
 
-    return (src: Observable<any>) =>
+    // inspired by https://medium.com/angular-in-depth/retry-failed-http-requests-in-angular-f5959d486294
+    return (src: Observable<AjaxResponse>) =>
         src.pipe(
             retryWhen(errors =>
                 errors.pipe(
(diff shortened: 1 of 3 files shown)


kilchenmann:
I will review it tomorrow. Just a short comment: title shoud have type "feat" and not "feature"

src/api/operators/retryOnError.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @param delayMs delay in milliseconds before the request is retried.
+ * @param maxRetries maximum number of retries.
+ * @param retryOnErrorStatus HTTP error status codes for which the request is retried.
+ * @param logError if true, error is written to the console error log.
+ *
+ * @category Internal
+ */
+export function retryOnError(delayMs: number, maxRetries: number, retryOnErrorStatus: number[], logError: boolean) {
+    let retries = maxRetries;
```
mpro7, line 17: Isn't it redundant? Why not directly use `maxRetries`?

Sorry, I wasn't asked for review, but was passing by and had a look ;)

mdelez (approved):
Works well! Nice operator :)

src/api/operators/retryOnError.ts
mdelez: I think it's because the number is modified below and it's not advisable to modify parameters directly.

tobiasschweizer: no worries :-)

I do this because on each retry the value is decremented (`retries--`). I generally try to avoid [reassigning to function params](https://eslint.org/docs/rules/no-param-reassign). However, I am sure there are a few place where I did this :-)

kilchenmann (approved):

PR #301, opened 2021-01-26
feat(ProjectMetadataEndpointV2): adjust x-n properties and custom converters (DSP-1222)

mpro7:
resolves DSP-1222
All properties which have cardinality `x-n` were adjusted and can be either single element or array of elements.


src/models/v2/custom-converters/union-advanced-url-object-converter.ts
```diff
@@ -0,0 +1,59 @@
+import { JsonConverter, JsonCustomConvert } from "json2typescript";
+import { Constants } from "../Constants";
+import { IUrl } from "./base-url-converter";
+
+/**
+ * @category Internal
+ */
+@JsonConverter
+export class UnionAdvancedUrlObjectConverter implements JsonCustomConvert<IUrl | string | [IUrl | string]> {
```
tobiasschweizer, line 9: same comment about tuples

src/models/v2/custom-converters/union-element-array-of-elements-converter.ts
```diff
@@ -0,0 +1,65 @@
...
+export class UnionElementArrayOfElementsConverter implements JsonCustomConvert<any> {
+
+    static jsonConvert: JsonConvert = new JsonConvert(
+        OperationMode.ENABLE,
+        ValueCheckingMode.DISALLOW_NULL,
+        false,
+        PropertyMatchingRule.CASE_STRICT
+    );
+    
+    serialize(el: any): any {
```
tobiasschweizer, line 22: couldn't annotate `el: any` as a specific type (either single element or array)?

src/models/v2/custom-converters/union-element-array-of-elements-converter.ts
```diff
@@ -0,0 +1,65 @@
...
+import { Attribution } from "../project-metadata/attribution";
+import { Grant } from "../project-metadata/grant";
+import { Place } from "../project-metadata/place";
+
+/**
+ * Experimental converter to distinguish elements and arrays of these as a first step of conversion
+ * @category Internal
+ */
+@JsonConverter
+export class UnionElementArrayOfElementsConverter implements JsonCustomConvert<any> {
```
tobiasschweizer, line 13: couldn't you use a more specific type than `any`?

src/models/v2/custom-converters/union-metadata-converter.ts
```diff
@@ -0,0 +1,55 @@
...
+import { PropertyMatchingRule } from "json2typescript/src/json2typescript/json-convert-enums";
+import { Constants } from "../Constants";
+import { Dataset } from "../project-metadata/dataset";
+import { Person } from "../project-metadata/person";
+
+/**
+ * @category Internal
+ */
+@JsonConverter
+export class UnionMetadataConverter implements JsonCustomConvert<[Dataset | Person]> {
```
tobiasschweizer, line 11: Would this also work `(Dataset | Person)[]`?

src/models/v2/project-metadata/attribution.ts
```diff
@@ -1,19 +1,22 @@
...
 /** 
  * @category Model V2 
  */ 
 @JsonObject("Attribution")
 export class Attribution extends BaseProjectMetadata {
 
-    @JsonProperty(Constants.DspHasRole, String)
-    role: string = "";
+    @JsonProperty(Constants.DspHasRole, UnionStringArrayOfStringsConverter)
+    role: string[] | string = [];
```
tobiasschweizer, line 16: could this always be an array? And then in the custom converter, you just deal with the JSON-LD representation which could be an array or a single object (both on reading and writing)?

tobiasschweizer (changes requested):
Please find some cursory comments that apply to various custom converters. I also commented on specific converters but I think these comments also apply to other custom converters.

## Consistent Use of Arrays

My suggestion is to always use arrays in the classes, and then in the custom converter, you just deal with the JSON-LD representation which could be an array or a single object (both on reading and writing).

Example: 

```ts
@JsonObject("Attribution")
export class Attribution extends BaseProjectMetadata {

@JsonProperty(Constants.DspHasRole, UnionStringArrayOfStringsConverter)
role: string[] | string = [];
```

Couldn't role (or roles) always be an array (`role: string[]`)? When reading from JSON-LD just always return an array, and when writing, turn an array with just one element into an object (see for example: https://github.com/dasch-swiss/dsp-js-lib/blob/main/src/models/v2/custom-converters/string-literal-to-string-literal-array-converter.ts).

I think this will make it easier to use the metadata endpoint's classes since the client always deals with the same structure.


##  Tuple vs. Array Annotation

I noticed that you use [tuple](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple) annotations. Could you use array annotations instead? I think the idea of tuples is that you know the number of elements which is not the case here I think.

This could be helpful: https://www.educative.io/edpresso/arrays-vs-tuples-in-typescript

src/models/v2/custom-converters/base-url-converter.ts
```diff
@@ -15,16 +15,28 @@ export interface IUrl {
...
-            [Constants.SchemaUrlValue]: val.value
+            "@type": el.type,
+            [Constants.SchemaUrlValue]: el.value
         };
     }
 
-    deserialize(val: any): IUrl {
+    serialize(el: IUrl): object | object[] {
+        if (Array.isArray(el)) {
+            const newObj = [] as any[];
```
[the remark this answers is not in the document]
mpro7: In multiple types arrays there is no other options to distinguish elements only with `son2typescript` library.

src/models/v2/custom-converters/base-url-converter.ts
```diff
@@ -15,16 +15,28 @@ export interface IUrl {
...
+            el.forEach((
+                item => newObj.push(this.serializeElement(item))
+            ));
+            return newObj;
+        } else {
+            return this.serializeElement(el);
+        }
+    }
+
+    deserialize(val: any): IUrl | IUrl[] | string | string[] {
```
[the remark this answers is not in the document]
mpro7: This is actually wrong, the type annonation should be `IUrl | string | (IUrl | string)[]` and as mentioned above, in multiple types array it isn't possible to use that approach.

src/models/v2/project-metadata/attribution.ts
mpro7: Good point, thanks for that. I'll  consider to change it in all possible properties, however this will require a lot of changes with the converters, so might end up moving this to another task.

src/models/v2/custom-converters/base-url-converter.ts
```diff
@@ -15,16 +15,28 @@ export interface IUrl {
...
         return {
-            "@type": val.type,
-            [Constants.SchemaUrlValue]: val.value
+            "@type": el.type,
+            [Constants.SchemaUrlValue]: el.value
         };
     }
 
-    deserialize(val: any): IUrl {
+    serialize(el: IUrl): object | object[] {
```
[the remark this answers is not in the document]
mpro7: Fixed in 2373bf0.

src/models/v2/custom-converters/base-url-converter.ts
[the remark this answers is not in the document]
mpro7: It could probably be, but as in `json2typescript` lib and in other converters we use any for JSON-LD, so let's keep it consistent. Map is not working here well.

src/models/v2/custom-converters/union-advanced-url-object-converter.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (el.hasOwnProperty(Constants.SchemaPropID)) {
+            const name = el[Constants.SchemaPropID][Constants.SchemaPropID];
+            const url = el[Constants.SchemaUrlValue];
+            return { name, url } as IUrl;
+        } else {
+            return el as string;
+        }
+    }
+
+    serialize(el: IUrl | string | [IUrl | string]): any {
```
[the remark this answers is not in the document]
mpro7: Fixed in 2373bf0.

src/models/v2/custom-converters/union-advanced-url-object-converter.ts
```diff
@@ -0,0 +1,59 @@
...
+            el.forEach((
+                item => newObj.push(this.serializeElement(item))
+            ));
+            return newObj;
+        } else {
+            return this.serializeElement(el);
+        }
+    }
+
+    deserialize(el: any): IUrl | string | [IUrl | string] {
```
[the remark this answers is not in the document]
mpro7: Fixed in 2373bf0.

src/models/v2/custom-converters/union-advanced-url-object-converter.ts
mpro7: Fixed in 2373bf0.

src/models/v2/custom-converters/union-element-array-of-elements-converter.ts
mpro7: Fixed in 2373bf0.

src/models/v2/custom-converters/union-element-array-of-elements-converter.ts
mpro7: Fixed in 2373bf0.

src/models/v2/custom-converters/union-metadata-converter.ts
mpro7: Fixed in 2373bf0.

src/models/v2/custom-converters/base-url-converter.ts
```diff
@@ -15,16 +15,28 @@ export interface IUrl {
  * @category Internal
  */
 @JsonConverter
-export abstract class BaseUrlConverter implements JsonCustomConvert<IUrl> {
+export abstract class BaseUrlConverter implements JsonCustomConvert<IUrl | IUrl[] | string | string[]> {
 
-    serialize(val: IUrl): object {
+    serializeElement(el: IUrl): object {
```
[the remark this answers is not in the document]
mpro7: In this case `protected` but good point, changed access modifiers in d362a3e.

src/models/v2/custom-converters/base-url-converter.ts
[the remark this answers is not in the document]
tobiasschweizer: Maybe we'll figure out why someday, but this is not a priority now. There must some reason related to type inference.

tobiasschweizer (approved):
I think it would be good to start adding some unit tests soon. E2E tests can easily break and it is sometimes not so simple to spot why they failed exactly.

mpro7:
> I think it would be good to start adding some unit tests soon. E2E tests can easily break and it is sometimes not so simple to spot why they failed exactly.

Thanks for review. Yeah, but how w/o test data?

tobiasschweizer:
> Yeah, but how w/o test data?

valid question

PR #302, opened 2021-01-26
fix(lists-endpoint): getListNodeInfo return types (DSP-1278)

mdelez:
resolves DSP-1278


tobiasschweizer (approved):
Looks good!

However, if the client should always get the same type of response, you could create a task for this. But this would change the API so maybe we better do not do it.

PR #304, opened 2021-01-27
feat(ontology endpoint)!: replace cardinalities (DSP-1265)

tobiasschweizer:
resolves DSP-1265

This PR introduces a breaking change for `OntologiesEndpointV2.addCardinalityToResourceClass`:

It now takes an argument of type `UpdateOntology<UpdateOntologyResourceClassCardinality>` which makes it consistent with the other methods that have a payload.


tobiasschweizer:
@kilchenmann We should coordinate the breaking change coming with this PR with https://github.com/dasch-swiss/dsp-js-lib/pull/210 which is also a breaking change.

kilchenmann:
> @kilchenmann We should coordinate the breaking change coming with this PR with #210 which is also a breaking change.

We should merge both breaking change PRs before we do a new release. This is not a problem, when it doesn't take too much time for PR #210

tobiasschweizer:
> > @kilchenmann We should coordinate the breaking change coming with this PR with #210 which is also a breaking change.
> 
> We should merge both breaking change PRs before we do a new release. This is not a problem, when it doesn't take too much time for PR #210

I still wonder wether it would make sense to still publish RCs, especially if they come with breaking changes. Publishing an RC still offers the possibility to rework things without having too many consequences.

Releases are not just a problem of automating processes but also of planning when which feature should be provided. As long as features are brand new, the could be published as RCs. But of course this only works if not everything is urgent ...

kilchenmann (approved):

PR #307, opened 2021-01-29
feat(permissions endpoint): Delete Permission (DSP-1304)

tobiasschweizer:
resolves DSP-1304

~~This PR requires DSP 13.1.0 (functionality and test data) which has not been released yet.~~


src/api/admin/permissions/permissions-endpoint.spec.ts
```diff
@@ -442,4 +443,50 @@ describe("PermissionsEndpoint", () => {
...
+                }
+            );
+
+            const request = jasmine.Ajax.requests.mostRecent();
+
+            const deleteDOAPermissionResponse = require("../../../../test/data/api/admin/permissions/delete-defaultObjectAccess-permission-response.json");
+
+            request.respondWith(MockAjaxCall.mockResponse(JSON.stringify(deleteDOAPermissionResponse)));
+
+            expect(request.url).toBe("http://localhost:3333/admin/permissions/http%3A%2F%2Frdfh.ch%2Fpermissions%2F00FF%2FDOAP-with-customIri");
```
waychal, line 484: Is it possible to remove hard coded localhost url from here? I noticed it also in other tests.

tobiasschweizer: This is because the connection is set up like this: `const config = new KnoraApiConfig("http", "localhost", 3333, undefined, undefined, true);`

Whatever is set up in the connection settings is reflected in the HTTP calls.

waychal: Ok.

tobiasschweizer: In fact, anything could be used in the setup. It is just important that the setting that was chosen in the setup is reflected in the HTTP call.

But maybe I did not understand your question correctly. What would be the alternative here?

waychal:
I do not see approve button anymore but everything else look fine to me.

tobiasschweizer:
> I do not see approve button anymore but everything else look fine to me.

Maybe you have to go to the tab "Files changed" and hit the button "Review changes". I wondered about this too.

waychal (approved):

waychal:
> > I do not see approve button anymore but everything else look fine to me.
> 
> Maybe you have to go to the tab "Files changed" and hit the button "Review changes". I wondered about this too.

Thanks.

subotic (approved):
LGTM!

PR #309, opened 2021-01-29
feat(metadata-endpoint): expose all x-n properties as arrays (DSP-1306)

mpro7:
resolves DSP-1306

### DIFF
--- a/test-framework/src/app/app.component.html
+++ b/test-framework/src/app/app.component.html
@@ -71,15 +71,6 @@
         </div>
     </section>
 
-    <section id="ontologyendpoint">
-        <h3>Ontology With All Languages</h3>
-        <button class="knora-api" (click)="getOntologyWithAllLanguages('http://api.knora.org/ontology/knora-api/v2')">Get system ontology</button> <br /><br />
-        <button class="anything" (click)="getOntologyWithAllLanguages('http://0.0.0.0:3333/ontology/0001/anything/v2')">Get anything ontology</button> <br /><br />
-        <div>
-            Ontology with all languages: <span class="ontology">{{ontologyWithAllLangs?.label}}</span>
-        </div>
-    </section>
-
     <section id="ontologymetadata">
         <h3>Project specific ontologies</h3>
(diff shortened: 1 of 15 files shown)


test-framework/src/app/app.component.html
```diff
@@ -177,9 +177,10 @@ <h3>Values</h3>
 
     <section id="metadata">
         <h3>Project Metadata</h3>
+        <button class="login" (click)="login()">Login</button><br /><br />
```
mdelez, line 180: I think this can be removed. We should only have one login button on the page.

test-framework/src/app/app.component.ts
```diff
@@ -1075,17 +1074,17 @@ export class AppComponent implements OnInit {
...
         //     'type': 'https://schema.org/URL',
         //     'value': 'http://p3.snf.ch/testproject'
         //  };
-        testDataset.project.grant = grant;
-        testDataset.project.keywords = 'science';
+        testDataset.project.grant = [grant];
+        testDataset.project.keywords = ['science'];
         testDataset.project.name = 'Testprojektname (test)';
-        testDataset.project.publication = 'testpublication';
+        testDataset.project.publication = ['testpublication']
```
mdelez, line 1087: missing semicolon

mdelez (changes requested):

test-framework/src/app/app.component.ts
mpro7: Added in e4f104a.

test-framework/src/app/app.component.html
mpro7: Normally I would agree, but since I actively develop metadata endpoint it really saves time to have another login button next to my buttons :) and no need to endless scroll up and down. I used to keep this change locally but got tired of remembering about that to not stage etc.

mdelez: ok, but we should remove it when everything for the metadata endpoint in js-lib is finished.

mdelez (approved):

test-framework/src/app/app.component.html
mpro7: That's the plan ;)

PR #310, opened 2021-01-29
feat(lists-endpoint): add support for deleting a list node(DSP-1049)

mdelez:
resolves DSP-1049


tobiasschweizer:
@mdelez Aye, will review today!

mdelez:
> @mdelez Aye, will review today!

Merci Monsieur Suisse!

src/api/admin/lists/lists-endpoint.spec.ts
```diff
@@ -525,4 +527,32 @@ describe("ListsEndpoint", () => {
...
+                (res: ApiResponseData<DeleteListNodeResponse | DeleteListResponse>) => {
+                    expect(res.body instanceof DeleteListNodeResponse).toBeTruthy();
+                    expect((res.body as DeleteListNodeResponse).node.id).toEqual("http://rdfh.ch/lists/0001/notUsedList");
+                    done();
+                }
+            );
+
+            const request = jasmine.Ajax.requests.mostRecent();
+
+            const listsResponse = require("../../../../test/data/api/admin/lists/delete-list-node-response.json");
```
tobiasschweizer, line 544: Could you provide a test case for the root node case?

src/api/admin/lists/lists-endpoint.spec.ts
```diff
@@ -525,4 +527,32 @@ describe("ListsEndpoint", () => {
 
     });
 
+    describe("Method deleteListNode", () => {
+
+        it("should delete a list child node", done => {
+
+            knoraApiConnection.admin.listsEndpoint.deleteListNode("http://rdfh.ch/lists/0001/notUsedList015").subscribe(
```
tobiasschweizer, line 534: It actually dos not matter, bit I saw that "notUsedList015" is contained in the response. Does the response show how the list looks like after removal of the deleted node?

src/api/admin/lists/lists-endpoint.spec.ts
tobiasschweizer: Could you also provide a spec in the e2e test?

src/models/admin/delete-list-node-response.ts
```diff
@@ -0,0 +1,19 @@
+import { JsonObject, JsonProperty } from "json2typescript";
+
+import { ListNode } from "./list-node";
+
+/**
+ * A response containing a list node.
+ *
+ * @category Model Admin
+ */
+@JsonObject("DeleteListNodeResponse")
```
tobiasschweizer, line 10: I think you have to add that class to `index.ts`.

src/models/admin/delete-list-response.ts
```diff
@@ -0,0 +1,23 @@
+import { JsonObject, JsonProperty } from "json2typescript";
+
+/**
+ * A response containing a deleted flag and iri.
+ *
+ * @category Model Admin
+ */
+@JsonObject("DeleteListNodeResponse")
```
tobiasschweizer, line 8: I think you have to add that class to `index.ts`.

tobiasschweizer (changes requested):
Please see my comments.

src/models/admin/delete-list-response.ts
mdelez: exported in f801307

src/models/admin/delete-list-node-response.ts
mdelez: exported in f801307

src/api/admin/lists/lists-endpoint.spec.ts
mdelez: oh yeah, you're right. The response will not contain the node that was just deleted. I've changed it to delete "notUsedList016" for clarity in dba9bee

src/api/admin/lists/lists-endpoint.spec.ts
mdelez: additional unit test added in dba9bee

mdelez: e2e tests added in 94ee63a

test-framework/src/app/app.component.ts
```diff
@@ -1366,11 +1369,31 @@ export class AppComponent implements OnInit {
...
+    }
+
+    deleteListChildNode(): void {
+        const listItemIri = 'http://rdfh.ch/lists/0001/notUsedList015';
 
-        
+        this.knoraApiConnection.admin.listsEndpoint.deleteListNode(listItemIri).subscribe(
+            (res: ApiResponseData<DeleteListNodeResponse>) => {
+                console.log(res);
+                this.listChildren = res.body.node.children.length;
```
tobiasschweizer, line 1383: I think it is good practice to add the error case as well: 

`err => console.error(err)`

This way, protractor will surely throw an error, see https://github.com/dasch-swiss/dsp-js-lib/blob/cf0d228722e805cbb38b9dff90115392482bebcd/test-framework/e2e/src/app.e2e-spec.ts#L910-L916

This should make the e2e tests fail even if there is no real assertion in the spec for the expected response.

I think this has not been done consistently so far, but I try to stick to this for new code ;-)

mdelez:
> This test is about deleting a root node, right?

it first deletes a child node and then the root node. I do it in the the same `it` because the order in which the tests run is random so the list could be deleted before deleting a child node in the list.

tobiasschweizer:
> > This test is about deleting a root node, right?
> 
> it first deletes a child node and then the root node. I do it in the same `it` because the order in which the tests run is random so the list could be deleted before deleting a child node in the list.

Ok, got it :-)

I think the e2e tests are actually always executed in the same order (unlike the unit tests). But you are right: there shouldn't be a dependency between the different `it`s. However, this would mean to reinit the repo after each `it.`

The have some dependency via the db's state: for example, sometimes a think is created, modified, and deleted thereafter.

tobiasschweizer (approved):
Looks good!

test-framework/src/app/app.component.ts
mdelez: done in 24b60c0

PR #312, opened 2021-02-01
feat(metadata-endpoint): test/implement flat response option (DSP-1223)

mpro7:
resolves DSP-1223

### DIFF
--- a/src/api/v2/project-metadata/metadata-endpoint-v2.ts
+++ b/src/api/v2/project-metadata/metadata-endpoint-v2.ts
@@ -6,7 +6,7 @@ import { ApiResponseError } from "../../../models/api-response-error";
 import { MetadataConversionUtil } from "../../../models/v2/project-metadata/metadata-conversion-util";
 import { ProjectsMetadata } from "../../../models/v2/project-metadata/project-metadata";
 import { UpdateProjectMetadataResponse } from "../../../models/v2/project-metadata/update-project-metadata";
-import { Endpoint } from "../../endpoint";
+import { Endpoint, IHeaderOptions } from "../../endpoint";
 
 declare let require: any; // http://stackoverflow.com/questions/34730010/angular2-5-minute-install-bug-require-is-not-defined
 const jsonld = require("jsonld/dist/jsonld.js");
@@ -26,7 +26,8 @@ export class ProjectMetadataEndpointV2 extends Endpoint {
      * @param resourceIri the Iri of the resource the value belongs to.
      */
     getProjectMetadata(resourceIri: string): Observable<ProjectsMetadata | ApiResponseError> {
-        return this.httpGet("/" + encodeURIComponent(resourceIri)).pipe(
+        const flatOption: IHeaderOptions = { "X-Knora-JSON-LD-Rendering": "flat" };
+        return this.httpGet("/" + encodeURIComponent(resourceIri), flatOption).pipe(
             // expand all Iris
             mergeMap((res: AjaxResponse) => {
                 return jsonld.compact(res.response, {});
--- a/src/index.ts
+++ b/src/index.ts
@@ -209,7 +209,6 @@ export { Grant } from "./models/v2/project-metadata/grant";
 export { Organization } from "./models/v2/project-metadata/organization";
 export { Place } from "./models/v2/project-metadata/place";
 export { UpdateProjectMetadataResponse } from "./models/v2/project-metadata/update-project-metadata";
-export { IUrl } from "./models/v2/custom-converters/base-url-converter";
 
 // Utils
 export { CardinalityUtil } from "./models/v2/resources/cardinality-util";
@@ -219,3 +218,5 @@ export { PermissionUtil } from "./models/v2/resources/permission-util";
 export { IPermissions } from "./interfaces/models/admin/i-permissions";
 export { IFulltextSearchParams } from "./interfaces/models/v2/i-fulltext-search-params";
 export { ILabelSearchParams } from "./interfaces/models/v2/i-label-search-params";
+export { IUrl } from "./models/v2/custom-converters/base-url-converter";
+export { IId } from "./models/v2/custom-converters/union-data-management-plan-id-converter";
--- a/src/index.ts
+++ b/src/index.ts
@@ -218,5 +218,4 @@ export { PermissionUtil } from "./models/v2/resources/permission-util";
 export { IPermissions } from "./interfaces/models/admin/i-permissions";
 export { IFulltextSearchParams } from "./interfaces/models/v2/i-fulltext-search-params";
 export { ILabelSearchParams } from "./interfaces/models/v2/i-label-search-params";
-export { IUrl } from "./models/v2/custom-converters/base-url-converter";
-export { IId } from "./models/v2/custom-converters/union-data-management-plan-id-converter";
+export { IId, IUrl } from "./models/v2/project-metadata/metadata-interfaces";
--- a/src/models/v2/custom-converters/union-data-management-plan-id-converter.ts
+++ b/src/models/v2/custom-converters/union-data-management-plan-id-converter.ts
@@ -3,11 +3,15 @@ import { PropertyMatchingRule } from "json2typescript/src/json2typescript/json-c
 import { Constants } from "../Constants";
 import { DataManagementPlan } from "../project-metadata/data-management-plan";
 
+export interface IId {
+    id: string;
+}
+
 /**
  * @category Internal
  */
 @JsonConverter
-export class UnionDataManagementPlanIdConverter implements JsonCustomConvert<DataManagementPlan | object> {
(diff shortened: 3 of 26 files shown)


tobiasschweizer:
Aye, will check this PR asap.

src/models/v2/custom-converters/union-element-array-of-elements-converter.ts
```diff
@@ -46,9 +56,18 @@ export class UnionElementArrayOfElementsConverter implements JsonCustomConvert
...
                     || el[0].hasOwnProperty(Constants.SchemaUrlValue):
                     return UnionElementArrayOfElementsConverter.jsonConvert.deserializeArray(el, Place);
+                case el[0].hasOwnProperty("@id"):
+                    const newArr = [] as Array<object>;
+                    if (Array.isArray(el)) {
+                        el.forEach((
+                            (item: any) => newArr.push({id: (item as { [index: string]: string })["@id"]})
+                        ));
+                    };
+                    return newArr as any[];
```
tobiasschweizer, line 66: Could you use the type that you are actually dealing with instead of `any[]`?

src/models/v2/custom-converters/union-element-array-of-elements-converter.ts
```diff
@@ -60,9 +79,14 @@ export class UnionElementArrayOfElementsConverter implements JsonCustomConvert
                 // below condition is temp solution for data sent from the test app, due to different structure
                     || el.hasOwnProperty(Constants.SchemaUrlValue):
                     return UnionElementArrayOfElementsConverter.jsonConvert.deserializeArray([el], Place);
+                case el.hasOwnProperty("@id"):
+                    return {
+                        id: (el as { [index: string]: string })["@id"]
```
tobiasschweizer, line 84: If you used an interface for the id object, you could reuse it here.

src/models/v2/project-metadata/project-metadata.ts
```diff
@@ -10,5 +10,5 @@ import { Person } from "./person";
 export class ProjectsMetadata {
 
     @JsonProperty("@graph", UnionMetadataConverter)
-    projectsMetadata: Array<Dataset | Person> = [];
+    projectsMetadata: Array<MetadataClasses> = [];
```
tobiasschweizer, line 13: `export type MetadataClasses = Dataset | SingleProject | Person | Organization | Grant | DataManagementPlan;`

So now all these types could occur on the upper level? Is this convenient to use for the client?

src/api/v2/project-metadata/metadata-endpoint-v2.ts
```diff
@@ -26,7 +26,8 @@ export class ProjectMetadataEndpointV2 extends Endpoint {
      * @param resourceIri the Iri of the resource the value belongs to.
      */
     getProjectMetadata(resourceIri: string): Observable<ProjectsMetadata | ApiResponseError> {
-        return this.httpGet("/" + encodeURIComponent(resourceIri)).pipe(
+        const flatOption: IHeaderOptions = { "X-Knora-JSON-LD-Rendering": "flat" };
```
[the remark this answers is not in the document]
tobiasschweizer: I think I just found it in the docs:

> ### Hierarchical vs. Flat JSON-LD

The client can choose between hierarchical and flat JSON-LD. In hierarchical
JSON-LD, entities with IRIs are inlined (nested) where they are used. If the
same entity is used in more than one place, it is inlined only once, and other
uses just refer to its IRI. In Knora's flat JSON-LD, all entities with IRIs are located
at the top level of the document (in a `@graph` if there is more than one of them).
This setting does not affect blank nodes, which are always inlined (unlike in standard
flat JSON-LD). Knora ontologies are always returned in the `flat` rendering; other kinds
of responses default to `hierarchical`. To use this setting, submit the HTTP header
`X-Knora-JSON-LD-Rendering` with the value `hierarchical` or `flat`.

src/models/v2/project-metadata/project-metadata.ts
tobiasschweizer: Or is this the price to pay for not having to implement additional logic to fin the inlined element if it is referred to in another place?

src/models/v2/custom-converters/union-metadata-converter.ts
```diff
@@ -17,41 +23,59 @@ export class UnionMetadataConverter implements JsonCustomConvert<Array<Dataset |
...
+        const newObj = [] as Array<MetadataClasses>;
         el.forEach((
             (item: any) => newObj.push(this.deserializeElement(item))
         ));
         return newObj;
     }
 
-    private serializeElement(el: Dataset | Person): any {
+    private serializeElement(el: MetadataClasses): any {
         if (el.hasOwnProperty("type") && el["type"] === Constants.DspDataset) {
```
tobiasschweizer, line 43: I think the `if else if ...` statement here is much clearer than "switch (true)". Could you maybe apply this everywhere?

tobiasschweizer (commented):
Please see my comments.

src/api/v2/project-metadata/metadata-endpoint-v2.ts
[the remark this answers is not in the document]
mpro7: If this what you find doesn't answer your question, talk to Ben :).

src/models/v2/project-metadata/project-metadata.ts
mpro7: You can easily have a look by running test-framework. This is one of two possible options and looks more consistent in contrast to nested option, where you actually never know what you get in and what out. This would be probably best to have as starter in first iteration. And it always can be revert. However I prefer the flat option, because pointless is to fight with JSON-LD which is provided by the backend.

src/models/v2/custom-converters/union-data-management-plan-id-converter.ts
```diff
@@ -0,0 +1,67 @@
+import { JsonConvert, JsonConverter, JsonCustomConvert, OperationMode, ValueCheckingMode } from "json2typescript";
+import { PropertyMatchingRule } from "json2typescript/src/json2typescript/json-convert-enums";
+import { Constants } from "../Constants";
+import { DataManagementPlan } from "../project-metadata/data-management-plan";
+
+/**
+ * @category Internal
+ */
+@JsonConverter
+export class UnionDataManagementPlanIdConverter implements JsonCustomConvert<DataManagementPlan | object> {
```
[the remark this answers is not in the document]
mpro7: Done in 064ba55.

src/models/v2/custom-converters/union-element-array-of-elements-converter.ts
```diff
@@ -19,7 +19,7 @@ export class UnionElementArrayOfElementsConverter implements JsonCustomConvert
         PropertyMatchingRule.CASE_STRICT
     );
     
-    serialize(el: Attribution[] | Place[] | Grant[]): any {
+    serialize(el: Attribution[] | Place[] | Grant[]| object): any {
```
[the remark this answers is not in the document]
mpro7: Done in 064ba55.

src/models/v2/custom-converters/union-element-array-of-elements-converter.ts
mpro7: Done in 064ba55.

src/models/v2/custom-converters/union-element-array-of-elements-converter.ts
mpro7: Done in 064ba55.

src/models/v2/custom-converters/union-metadata-converter.ts
mpro7: I leave it as it is now and think about it while doing some refactor of metadata cute corner.

src/index.ts
```diff
@@ -219,3 +218,5 @@ export { PermissionUtil } from "./models/v2/resources/permission-util";
 export { IPermissions } from "./interfaces/models/admin/i-permissions";
 export { IFulltextSearchParams } from "./interfaces/models/v2/i-fulltext-search-params";
 export { ILabelSearchParams } from "./interfaces/models/v2/i-label-search-params";
+export { IUrl } from "./models/v2/custom-converters/base-url-converter";
```
tobiasschweizer, line 221: Maybe it would be better to define these interfaces elsewhere (not in the custom converter) since they are exposes to the lib's public API (whereas the custom converters are not).

tobiasschweizer (approved):

src/index.ts
mpro7: Yes, was thinking about that. Will do that during refactor task. Thanks for review.

tobiasschweizer: As long as the interface is called the same and has the same shape, moving them into another place won't affect usage from the outside.

mpro7: I know :) but as you see I already done it in 4b21374.

PR #322, opened 2021-02-23
refactor(create-child-node-request): add optional position property (DSP-1300)

mdelez:
resolves DSP-1300

### DIFF
--- a/src/api/admin/lists/lists-endpoint.spec.ts
+++ b/src/api/admin/lists/lists-endpoint.spec.ts
@@ -382,6 +382,7 @@ describe("ListsEndpoint", () => {
 
             knoraApiConnection.admin.listsEndpoint.createChildNode(childNode).subscribe(
                 (res: ApiResponseData<ListNodeInfoResponse>) => {
+                    expect(res.body.nodeinfo.id).toEqual('http://rdfh.ch/lists/0001/treeList01');
                     done();
                 }
             );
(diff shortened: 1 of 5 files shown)


test-framework/e2e/src/app.e2e-spec.ts
```diff
@@ -972,7 +1030,7 @@ describe('workspace-project App', () => {
 
     const listNodePosition = page.getEle('div#lists span.list-node-position');
 
-    expect(listNodePosition.getText()).toEqual('2');
+    expect(listNodePosition.getText()).toEqual('4');
```
tobiasschweizer, line 1033: Why is this 4 now?

tobiasschweizer (commented):
just two comments ;-)

test-framework/e2e/src/app.e2e-spec.ts
mdelez: Because I add 2 child nodes in the tests above. These tests seems to run in order. Same reason why I have to test the delete method as the last test, otherwise the list is gone and the other tests fail.

tobiasschweizer: Understood. What we are doing here is an anti pattern ;-), I found out reading the [cypress guidelines ](https://docs.cypress.io/guides/references/best-practices.html#Having-tests-rely-on-the-state-of-previous-tests)

tobiasschweizer: good that we are using protractor and not cypress

src/api/admin/lists/lists-endpoint.spec.ts
```diff
@@ -407,6 +407,55 @@ describe("ListsEndpoint", () => {
...
+            const comment1 = new StringLiteral();
+            comment1.language = "en";
+            comment1.value = "Inserted List Node Comment";
+
+            childNode.comments = [comment1];
+
+            childNode.position = 1;
+
+            knoraApiConnection.admin.listsEndpoint.createChildNode(childNode).subscribe(
+                (res: ApiResponseData<ListNodeInfoResponse>) => {
```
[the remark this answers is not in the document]
mdelez: added in 8112736 and 4f14605

tobiasschweizer (approved):

PR #324, opened 2021-02-25
feat(lists-endpoint): Support for receiving a list child node from DSP-API (DSP-1385)

mdelez:
resolves DSP-1385

### DIFF
--- a/src/api/admin/lists/lists-endpoint-admin.ts
+++ b/src/api/admin/lists/lists-endpoint-admin.ts
@@ -5,8 +5,10 @@ import { CreateChildNodeRequest } from "../../../models/admin/create-child-node-
 import { CreateListRequest } from "../../../models/admin/create-list-request";
 import { DeleteListNodeResponse } from "../../../models/admin/delete-list-node-response";
 import { DeleteListResponse } from "../../../models/admin/delete-list-response";
+import { ListChildNodeResponse } from "../../../models/admin/list-child-node-response";
 import { ListInfoResponse } from "../../../models/admin/list-info-response";
 import { ListNodeInfoResponse } from "../../../models/admin/list-node-info-response";
+import { ListNodeResponse } from "../../../models/admin/list-node-response";
 import { ListResponse } from "../../../models/admin/list-response";
 import { ListsResponse } from "../../../models/admin/lists-response";
 import { RepositionChildNodeRequest } from "../../../models/admin/reposition-child-node-request";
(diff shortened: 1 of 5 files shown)


mdelez:
@tobiasschweizer I'll discuss this during the meeting today

src/api/admin/lists/lists-endpoint-admin.ts
```diff
@@ -232,10 +234,16 @@ export class ListsEndpointAdmin extends Endpoint {
      * 
      * @param listItemIri The IRI of the list.
      */
-    getList(listItemIri: string): Observable<ApiResponseData<ListResponse> | ApiResponseError> {
+    getList(listItemIri: string): Observable<ApiResponseData<ListResponse | ListChildNodeResponse> | ApiResponseError> {
```
tobiasschweizer, line 237: Will this be regarded a breaking change?

tobiasschweizer: Not sure if code actively using `getList` already will still compile.

src/api/admin/lists/lists-endpoint.spec.ts
```diff
@@ -556,11 +557,11 @@ describe("ListsEndpoint", () => {
...
     describe("Method getList", () => {
 
-        it("should return a list", done => {
+        it("should return a list (root node)", done => {
 
             knoraApiConnection.admin.listsEndpoint.getList("http://rdfh.ch/lists/0001/treeList").subscribe(
-                (res: ApiResponseData<ListResponse>) => {
-                    expect(res.body.list.listinfo.id).toEqual("http://rdfh.ch/lists/0001/treeList");
+                (res: ApiResponseData<ListResponse | ListChildNodeResponse>) => {
+                    expect((res.body as ListResponse).list.listinfo.id).toEqual("http://rdfh.ch/lists/0001/treeList");
```
tobiasschweizer, line 564: I think without the assertion ` as ListResponse` this would not compile, right?

mdelez: It seems to work without the casts as well. I removed them in 0e40bb2

src/api/admin/lists/lists-endpoint-admin.ts
mdelez: I think if you only specify one return type, then it shouldn't break anything. I'm actually doing that in [this file ](https://github.com/dasch-swiss/dsp-app/blob/8ed2d9b0d61b64c0cde83d227b13c7614470e976/src/app/project/list/list-item/list-item.component.ts#L47-L51) and [below in the same file](https://github.com/dasch-swiss/dsp-app/blob/8ed2d9b0d61b64c0cde83d227b13c7614470e976/src/app/project/list/list-item/list-item.component.ts#L102-L111) I use the two different return types (currently testing with yalc)

tobiasschweizer (approved):

PR #328, opened 2021-03-01
fix(metadata-endpoint): types inconsistency (DSP-1388)

mpro7:
resolves DSP-1388


mpro7:
@waychal this should fixes all listed issues in the task. You can double check it by switching to my branch `wip/DSP-1388` and running DSP-JS from there.

waychal (approved):
Looks good to me.

PR #343, opened 2021-05-10
feat(ontology): new method to change gui order (DSP-1566)

kilchenmann:
resolves DSP-1566


mdelez (changes requested):
Works well for me! I would suggest adding this functionality to the app playground to keep it up to date.

PR #350, opened 2021-07-01
feat(ontology): new method to update gui element and attributes (DSP-1768)

kilchenmann:
resolves DSP-1768


kilchenmann:
I do not have test date from the API but maybe I will add the functionality to the test-framework and add a simple e2e test

PR #354, opened 2021-07-14
chore(models): expand resource conversion (DSP-1795)

kilchenmann:
resolves DSP-1795

Adds the missing conversion for file values like document, audio and moving image. I wanted to add some tests, but we do not have proper test-data for those cases.


kilchenmann:
There was a wrong indention and the vscode formatter fixed it in 78bb6eb0. The main changes were done in 307a56ca

subotic (approved):
LGTM, thanks!

PR #357, opened 2021-07-19
feat(resource): add erase resource method (DSP-1807)

kilchenmann:
resolves DSP-1807

### DIFF
--- a/src/api/v2/resource/resources-endpoint-v2.ts
+++ b/src/api/v2/resource/resources-endpoint-v2.ts
@@ -178,6 +178,29 @@ export class ResourcesEndpointV2 extends Endpoint {
             catchError(error => this.handleError(error))
         );
 
+    }
+
+    /**
+     * Erases a resource.
+     *
+     * @param resource the resource to be deleted.
+     */
+    eraseResource(resource: DeleteResource): Observable<DeleteResourceResponse | ApiResponseError> {
+
+        const res = this.jsonConvert.serializeObject(resource);
+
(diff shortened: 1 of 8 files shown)


src/api/v2/resource/resources-endpoint-v2.ts
```diff
@@ -178,6 +178,29 @@ export class ResourcesEndpointV2 extends Endpoint {
...
         );
 
+    }
+
+    /**
+     * Erases a resource.
+     *
+     * @param resource the resource to be deleted.
+     */
+    eraseResource(resource: DeleteResource): Observable<DeleteResourceResponse | ApiResponseError> {
```
mdelez, line 188: would "deleteResource" be a better name for this method? I don't think we use the word "erase" anywhere

kilchenmann: `deleteResource` method exists already and uses the `/delete` route of the API. `eraseResource` method is different from `deleteResource` and uses the `/erase` route of the API. This is why I choose the same name for the method.
It's all described here: <https://docs.dasch.swiss/DSP-API/03-apis/api-v2/editing-resources/#deleting-a-resource> and here <https://docs.dasch.swiss/DSP-API/03-apis/api-v2/editing-resources/#erasing-a-resource-from-the-triplestore>

mdelez: ah, I see :)

mdelez (approved):

PR #361, opened 2021-07-28
fix(endpoint)!: resolve set-cookie issue (DSP-1827)

kilchenmann:
resolves DSP-1827

### DIFF
--- a/src/api/endpoint.spec.ts
+++ b/src/api/endpoint.spec.ts
@@ -116,8 +116,7 @@ describe("Test class Endpoint", () => {
 
         expect(request.method).toEqual("GET");
 
-        // not needed anymore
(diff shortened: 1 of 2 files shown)


kilchenmann:
This is how it works now:

kilchenmann:
btw. for the reviewers: if you want to test it with the test-framework you have to open it on the same domain as the api is running e.g. http://0.0.0.0:4200 and you should see the cookie in the request headers if it was set on login.

Screen Shot 2021-07-28 at 12 14 30

subotic (approved):
Great, thanks! LGTM

kilchenmann:
> Edit: Or we could set crossDomain to true? Not sure if this works. I'll try after lunch....

No, crossDomain does not work. Not sure if it's supported from the API-side!?

mdelez (approved):
works well!

PR #364, opened 2021-08-03
chore(deps)!: bump DSP-API to v14.0.1

irinaschubert:
resolves DSP-1841


src/api/v2/resource/resources-endpoint.spec.ts
```diff
@@ -514,31 +516,31 @@ describe("ResourcesEndpoint", () => {
...
-            request.respondWith(MockAjaxCall.mockResponse(JSON.stringify(updateResResponse)));
+            request.respondWith(MockAjaxCall.mockResponse(JSON.stringify(updateResResponseWithLastModificationDate)));
 
             expect(request.url).toBe("http://0.0.0.0:3333/v2/resources");
 
             expect(request.method).toEqual("PUT");
 
             const expectedPayload = require("../../../../test/data/api/v2/resources/update-resource-metadata-request-with-last-mod-date-expanded.json");
 
-            // TODO: remove this bad hack once test data is stable
```
irinaschubert, line 538: should we leave this "bad hack" here or delete it? if we delete it, the modification dates have to be adjusted in the test after every release of dsp-api as new test data is generated with updated values.

src/api/v2/resource/resources-endpoint.spec.ts
```diff
@@ -468,9 +473,6 @@ describe("ResourcesEndpoint", () => {
 
             const expectedPayload = require("../../../../test/data/api/v2/resources/update-resource-metadata-request-expanded.json");
 
-            // TODO: remove this bad hack once test data is stable
```
irinaschubert, line 471: should we leave this "bad hack" here or delete it? if we delete it, the modification dates have to be adjusted in the test after every release of dsp-api as new test data is generated with updated values.

kilchenmann: I think it's fine to not test the modification date.

src/api/v2/resource/resources-endpoint.spec.ts
kilchenmann: I think it's fine to not test the modification date.

kilchenmann (commented):
You have also to commit all the test data. 
And when I run `npm i` my package-lock.json file changed. So, there's something wrong. In my case it changes the version of "karma-typescript" back to "^5.0.3". Have you changed something in the package.json? Then you have to commit this too.

src/api/v2/resource/resources-endpoint.spec.ts
irinaschubert: OK. Then I suggest setting the dates manually like it was before. I just changed the description as in my opinion there is  no "problem" with unstable test data. Modification dates of resources are never stable by definition. But we don't want to adapt these tests after every release of dsp-api. So, this is a good solution, I think.

src/api/v2/resource/resources-endpoint.spec.ts
irinaschubert: same as above

kilchenmann:
Btw the title should be changed. Because the PR is about the update to the API dependency. Normally we write it like this: "chore(deps): bump DSP-API to v14.0.1"
This helps to see in the release notes directly in which version we updated the API.
I'm not sure if we have to mark this PR as a breaking change? Because we update the API to a new major version. Otherwise the next version (after this PR has been merged) will be 3.0.1.
What do you think @irinaschubert?

irinaschubert:
> You have also to commit all the test data.
> And when I run `npm i` my package-lock.json file changed. So, there's something wrong. In my case it changes the version of "karma-typescript" back to "^5.0.3". Have you changed something in the package.json? Then you have to commit this too.

I have no idea what I did there exactly... I installed some npm packages in order to be able to run the tests and I guess the package.json was then updated automatically. I committed it now. I committed also the test data.

irinaschubert:
> Btw the title should be changed. Because the PR is about the update to the API dependency. Normally we write it like this: "chore(deps): bump DSP-API to v14.0.1"
> This helps to see in the release notes directly in which version we updated the API.
> I'm not sure if we have to mark this PR as a breaking change? Because we update the API to a new major version. Otherwise the next version (after this PR has been merged) will be 3.0.1.
> What do you think @irinaschubert?

I agree with the PR title and changed it. I'm not sure what to say about your second question: the PR does not contain breaking changes, does it? Is the dsp-js-lib versioning linked to the versioning of dsp-api?

kilchenmann:
> > Btw the title should be changed. Because the PR is about the update to the API dependency. Normally we write it like this: "chore(deps): bump DSP-API to v14.0.1"
> > This helps to see in the release notes directly in which version we updated the API.
> > I'm not sure if we have to mark this PR as a breaking change? Because we update the API to a new major version. Otherwise the next version (after this PR has been merged) will be 3.0.1.
> > What do you think @irinaschubert?
> 
> I agree with the PR title and changed it. I'm not sure what to say about your second question: the PR does not contain breaking changes, does it? Is the dsp-js-lib versioning linked to the versioning of dsp-api?

I just had a look at <https://github.com/dasch-swiss/dsp-js-lib/pull/364/files#diff-1feab4f64ae74d36a0833132308aa492e93bf84acc511f44208b01bf6fda6aff> where we have a breaking change. The response of the resource metadata update has changed. So, in fact we have a breaking change. 

p.s. I wasn't sure before, but now I think we have to mark is a breaking change with `chore(deps)!: bump DSP-API to v14.0.1`

irinaschubert:
> > > Btw the title should be changed. Because the PR is about the update to the API dependency. Normally we write it like this: "chore(deps): bump DSP-API to v14.0.1"
> > > This helps to see in the release notes directly in which version we updated the API.
> > > I'm not sure if we have to mark this PR as a breaking change? Because we update the API to a new major version. Otherwise the next version (after this PR has been merged) will be 3.0.1.
> > > What do you think @irinaschubert?
> > 
> > 
> > I agree with the PR title and changed it. I'm not sure what to say about your second question: the PR does not contain breaking changes, does it? Is the dsp-js-lib versioning linked to the versioning of dsp-api?
> 
> I just had a look at https://github.com/dasch-swiss/dsp-js-lib/pull/364/files#diff-1feab4f64ae74d36a0833132308aa492e93bf84acc511f44208b01bf6fda6aff where we have a breaking change. The response of the resource metadata update has changed. So, in fact we have a breaking change.
> 
> p.s. I wasn't sure before, but now I think we have to mark is a breaking change with `chore(deps)!: bump DSP-API to v14.0.1`

OK, I believe you although I don't quite understand it: the change you mention is a change in dsp-api not in js-lib, right? So, we only had to update the tests and nothing else in js-lib... I think I still miss a piece to fully understand the versioning here.

kilchenmann:
> > > > Btw the title should be changed. Because the PR is about the update to the API dependency. Normally we write it like this: "chore(deps): bump DSP-API to v14.0.1"
> > > > This helps to see in the release notes directly in which version we updated the API.
> > > > I'm not sure if we have to mark this PR as a breaking change? Because we update the API to a new major version. Otherwise the next version (after this PR has been merged) will be 3.0.1.
> > > > What do you think @irinaschubert?
> > > 
> > > 
> > > I agree with the PR title and changed it. I'm not sure what to say about your second question: the PR does not contain breaking changes, does it? Is the dsp-js-lib versioning linked to the versioning of dsp-api?
> > 
> > 
> > I just had a look at https://github.com/dasch-swiss/dsp-js-lib/pull/364/files#diff-1feab4f64ae74d36a0833132308aa492e93bf84acc511f44208b01bf6fda6aff where we have a breaking change. The response of the resource metadata update has changed. So, in fact we have a breaking change.
> > p.s. I wasn't sure before, but now I think we have to mark is a breaking change with `chore(deps)!: bump DSP-API to v14.0.1`
> 
> OK, I believe you although I don't quite understand it: the change you mention is a change in dsp-api not in js-lib, right? So, we only had to update the tests and nothing else in js-lib... I think I still miss a piece to fully understand the versioning here.

The breaking change in the API has also an effect in JS-Lib: the object class `UpdateResourceMetadataResponse` has changed. Because it does not contain only a string anymore. As a developer who uses the JS-Lib I have to know about this change.

kilchenmann (approved):

PR #365, opened 2021-08-09
add canDeleteCardinalitiesFromClass / deleteCardinalitiesFromClass endpoints

subotic:
resolves DSP-1700

### DIFF
--- a/src/api/v2/ontology/ontologies-endpoint-v2.ts
+++ b/src/api/v2/ontology/ontologies-endpoint-v2.ts
@@ -588,6 +588,70 @@ export class OntologiesEndpointV2 extends Endpoint {
     }
 
     /**
+     * Checks wether a cardinality can be removed from a class even in presence of resources of this class.
+     * A cardinality can be removed in the case that the property is not used in any resources.
+     * For now, DSP-API allows only one cardinality at a time to delete.
+     * @param deleteCardinalitiesFromClass the cardinalities that need to be checked.
+     */
+    canDeleteCardinalitiesFromClass(deleteCardinalitiesFromClass: UpdateOntology<UpdateResourceClassCardinality>): Observable<CanDoResponse | ApiResponseError> {
+        const deleteCardinalitiesFromClassRequest = this.jsonConvert.serializeObject(deleteCardinalitiesFromClass);
+
+        const numberOfCardinalities = deleteCardinalitiesFromClass.entity.cardinalities.length
+
+        if (numberOfCardinalities > 1) {
+            // FIXME: Return error - currently only one cardinality at a time can be deleted
+        }
+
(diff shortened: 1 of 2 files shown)


src/api/v2/ontology/ontologies-endpoint-v2.ts
```diff
@@ -587,6 +587,70 @@ export class OntologiesEndpointV2 extends Endpoint {
 
     }
 
+    /**
+     * Checks wether a cardinality can be removed from a class even in presence of resources of this class.
+     * A cardinality can be removed in the case that the property is not used in any resources.
+     * For now, DSP-API allows only one cardinality at a time to delete.
+     * @param deleteCardinalitiesFromClass the cardinalities that need to be checked.
+     */
+    canDeleteCardinalitiesFromClass(deleteCardinalitiesFromClass: UpdateOntology<UpdateResourceClassCardinality>): Observable<CanDoResponse | ApiResponseError> {
```
kilchenmann, line 596: may I rename to `canDeleteCardinalityFromResourceClass`? We should use the singular term of cardinalities, because the request is only for one cardinality, right? And ResourceClass instead Class only, to keep it consistent.
Same for `deleteCardinalityFromResourceClass`

subotic: sure

src/api/endpoint.ts
```diff
@@ -140,11 +140,11 @@ export class Endpoint {
      * @param path the relative URL for the request.
      * @param headerOpts additional headers, if any.
      */
-    protected httpDelete(path?: string, headerOpts?: IHeaderOptions): Observable<AjaxResponse> {
+    protected httpDelete(path?: string, body?: any, contentType: "json" | "sparql" = "json", headerOpts?: IHeaderOptions): Observable<AjaxResponse> {
 
         if (path === undefined) path = "";
 
-        return ajax(this.setAjaxRequest(path, "DELETE", undefined, this.constructHeader(undefined, headerOpts)))
+        return ajax(this.setAjaxRequest(path, "DELETE", body, this.constructHeader(contentType, headerOpts)))
```
kilchenmann, line 147: The `AjaxRequest` does not allow to have a body within DELETE method.

Screen Shot 2021-09-07 at 10 18 21

This is why the tests always fail.
@subotic any suggestion how to solve this issue?

subotic: Not very nice of rxJS.  I can change the implementation to a PUT or POST. If those are already used, then we have PATCH which also takes a body.

kilchenmann: Hi @subotic. I thought your in vacation 🏝 😉 
Ehm, I already talked to @irinaschubert about this topic. We will have a look at it. No worries.

kilchenmann:
I just integrated the canDelete- and deleteCardinality in a separate PR (#372) which should close this one without merging.

PR #367, opened 2021-08-09
feat(ontologies): new method about delete cardinalities

kilchenmann:
resolves DSP-1700 (--> TODO: this PR will get its own task on youtrack)


subotic:
@kilchenmann what about this PR? https://github.com/dasch-swiss/dsp-js-lib/pull/365 I have already added those two methods.

PR #386, opened 2021-12-09
chore(project-metadata): remove project metadata implementation (DEV-282)

mdelez:
resolves DEV-282


mdelez:
@mpro7 I'm not sure if everything related to `IUrl` is part of the metadata. Do you know if it can also be removed?

mpro7:
@mdelez as far as I remember I've created it while adding metadata related stuff and it was only used there.

mdelez:
@mpro7 ok, then I will remove it

mpro7:
@mdelez you can also remove (unless you will find some usage):
`src/interfaces/models/v2/project-metadata-interfaces.ts`
and from below location
`../models/v2/custom-converters/`
- all `union-` converters
- distribution-converter
- place-converter
- id-converter

mpro7 (approved):
Ok I think you already did what I've asked in prev comment.

PR #396, opened 2022-01-26
feat(constants): add missing constants

kilchenmann:
To avoid a construct like this in DSP-App `subProp === Constants.KnoraApiV2 + Constants.HashDelimiter + 'isPartOf'`, I've updated the constants in DSP-JS.


mdelez (approved):
Thanks!

PR #401, opened 2022-01-30
chore(gh): update gh-actions workflow

kilchenmann:
Similar workflow as in DSP-APP. Not matrix environment needed


mdelez (approved):
Thanks!

PR #462, opened 2022-10-06
chore: maintain dependencies

mpro7:
resolves DEV-1388


.github/workflows/main.yml
```diff
@@ -1,6 +1,6 @@
 name: CI
 env:
-  NodeVersion: 12.x  # node version
+  NodeVersion: 16.15.0
```
mdelez, line 3: should this be specifically 16.15.0?

mdelez (approved):

.github/workflows/main.yml
mpro7: The same is in APP now. It could be changed as soon as the newest version is not causing the `peer-deps` errors on GH CI - which was the case I decided to set this instead of `16.x`.

PR #485, opened 2023-01-17
feat(config): extend api config to use a second port (DEV-1609)

mdelez:
resolves DEV-1609


src/api/admin/projects/projects-endpoint-admin.ts
```diff
@@ -24,9 +24,10 @@ export class ProjectsEndpointAdmin extends Endpoint {
     /**
      * Returns a list of all projects.
      */
-    getProjects(): Observable<ApiResponseData<ProjectsResponse> | ApiResponseError> {
-    
-        return this.httpGet("").pipe(
+    getProjects(zioRequest?: boolean): Observable<ApiResponseData<ProjectsResponse> | ApiResponseError> {
+        if (zioRequest === undefined) zioRequest = false;
```
mpro7, line 28: If it will not more harder to implement it also on the APP side I would op to make it required and with default value  as false, so if statement isn't necessary anymore:

```suggestion
    getProjects(zioRequest: boolean = false): Observable<ApiResponseData<ProjectsResponse> | ApiResponseError> {
```

You could also add the param to the docstring.

This suggestion IMO applies everywhere `zioRequest` us used.

src/knora-api-config.ts
```diff
@@ -33,6 +42,7 @@ export class KnoraApiConfig {
     constructor(public apiProtocol: "http" | "https",
                 public apiHost: string,
                 public apiPort: number | null = null,
+                public zioApiPort: number | null = null,
```
mpro7, line 45: I'm not sure if I get all this correctly, but maybe here also default value would do the job, so we don't even need extra `zioApiPort` param, but we use `apiPort` with default at 3333. Let me know if this actually makes sense.

mpro7 (commented):
Check my comments please.

src/api/admin/projects/projects-endpoint-admin.ts
mdelez: done in [f609cd7](https://github.com/dasch-swiss/dsp-js-lib/pull/485/commits/f609cd73b9e312532c23e6e6024178fac2e93850)

src/knora-api-config.ts
mdelez: as discussed, this will take some time to implement and test so we will do this in another PR

mpro7 (approved):

PR #495, opened 2023-02-06
feat!(zio-http): additional zio http support (DEV-1705)

mdelez:
resolves DEV-1705

### DIFF
--- a/test-framework/src/app/app.component.ts
+++ b/test-framework/src/app/app.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { Project, ProjectResponse, ProjectsResponse } from '@dasch-swiss/dsp-js';
+import { KeywordsResponse, MembersResponse, Project, ProjectResponse, ProjectRestrictedViewSettingsResponse, ProjectsResponse, UpdateProjectRequest } from '@dasch-swiss/dsp-js';
 import {
     AdministrativePermissionResponse,
     AdministrativePermissionsResponse,
@@ -157,7 +157,14 @@ export class AppComponent implements OnInit {
     listNodePosition = 0;
     listNodeParentIri = '';
 
+
+    allProjectsKeywordsCount = 0;
+
     projectCount = 0;
+    projectInfo = '';
+    projectKeywordsCount = 0;
+    projectMembersCount = 0;
+    projectAdminMembersCount = 0;
 
     // canDoResponse: boolean;
 
@@ -169,7 +176,7 @@ export class AppComponent implements OnInit {
             undefined,
             undefined,
             true,
-            ":5555",
+            ':5555',
             ['/admin/projects']);
         this.knoraApiConnection = new KnoraApiConnection(config);
         // console.log(this.knoraApiConnection);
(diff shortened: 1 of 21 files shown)


BalduinLandolt (approved):
Looks good, and in any case much cleaner than the boolean solution.

test-framework/src/app/app.component.ts
```diff
@@ -162,7 +162,15 @@ export class AppComponent implements OnInit {
...
-        const config = new KnoraApiConfig('http', '0.0.0.0', 3333, 5555, undefined, undefined, true);
+        const config = new KnoraApiConfig(
+            'http',
+            '0.0.0.0',
+            3333,
+            undefined,
+            undefined,
+            true,
+            ":5555",
+            ['/admin/projects']);
```
mpro7, line 173: Does the new concept assume that all methods under this endpoint should be ziofied? What if some of them will be not (like they implemented in chunks)? This is more question to @BalduinLandolt I guess.

mpro7 (approved):
Looks good, apart of one unknown:

test-framework/src/app/app.component.ts
BalduinLandolt: correct. I was under the assumption that this is good enough for us, and it's less work and less config required in the App/Js-Lib. But if you think it should be more fine-grained, you'd have to see with @mdelez how much extra work this causes him.

mpro7: You've omitted the answer for my question, what if some methods are not ziofied? Would this still work? Other than that I like this idea, but we need to make sure that this will not break API.

BalduinLandolt: No, this requires all routes under the specified endpoint(s) to be ziofied. Meaning, we can not enable an endpoint to use the new zio routes, unless all routes under this endpoint are ziofied. But since we leave the old akka routes in, we can simply not enable an endpoint, while not all routes are finished, so if we stick to this rule, nothing should break.

mpro7: So final question, are project routes ready to be enabled? I guess we are aiming here in the next release as it will push this changes to production then.

BalduinLandolt: from what I understand, they are ready. And linear suggests the same: https://linear.app/dasch/issue/DEV-1587/dsp-api-expose-adminprojects-routes-with-zio-http

But I suppose this PR should be tested on DEV before pushing it to production. But this in turn opens the can of worms of releasing pre-releases of js-lib to npm, right?

mpro7: You're right, it should be tested this way and it's doable, but our setup...well I spoke with @mdelez about that, so he/we will handle it before the release. Still one+ week to go.

mdelez: We also need to coordinate with Sami to set up the redirect rules on the server, otherwise the routes won't work

mdelez: Redirect rules have been set up in this PR https://github.com/dasch-swiss/ops-deploy/pull/345

PR #499, opened 2023-02-15
feat(ontologies-endpoint): add new cardinalities check route (DEV-1704)

mdelez:
resolves DEV-1704


src/models/v2/resources/cardinality-util.ts
```diff
@@ -24,6 +24,19 @@ export namespace CardinalityUtil {
...
+                return "0-1"
+            case Cardinality._0_n:
+                return "0-n"
+            case Cardinality._1:
+                return "1"
+            case Cardinality._1_n:
+                return "1-n"
+        }
+    }
+
```
mpro7, line 39: Wouldn't be nicer to make a Map<Cardinality, string> a const and then `.get` the value or `for...of` the keys, just instead of 2 switches.

test-framework/e2e/src/app.e2e-spec.ts
```diff
@@ -776,8 +776,96 @@ describe('workspace-project App', () => {
     const getButton = page.getEle('div section#ontologyeditor button.read-onto');
     getButton.click();
 
-    const canDelete = page.getEle('div button.can-replace-card-for-res-card');
-    canDelete.click();
```
mpro7, line 780: Just to confirm this is not needed anymore?

mpro7 (approved):
Looks good. If both question I asked below are no === merge ;)

test-framework/e2e/src/app.e2e-spec.ts
mdelez: this was incorrectly named probably due to copy-pasting

src/models/v2/resources/cardinality-util.ts
mdelez: Doing this would make it more flexible but I'm not sure if it's really needed. I have the second switch in the `app` component because I don't have access to `Cardinality` in the template to pass it into the method. I'm not sure if this would ever actually be an issue in the DSP-APP. Having said that, I can try it out and see how it works.

mdelez: done in [8b2406c](https://github.com/dasch-swiss/dsp-js-lib/pull/499/commits/8b2406c84d69bf2d1bb4f3b37c2ae4a1f09c3ec3)

PR #504, opened 2023-02-20
feat(package.json): modify and add new yalc npm scripts

mdelez:
Our npm script to publish the package via yalc has the `--push` flag which automatically updates any local repository using the library to this new version (as seen in the image below). This is sometimes not ideal as you don't want to try this new version out in the DSP-APP but only the test app within the JS-LIB repo. I've modified the npm script to only publish the package and renamed the old npm script to something clearer.

Screen Shot 2023-02-20 at 09 28 55


mpro7 (approved):
`yalc-publish` is also used to run in e2e test but I guess if they passed on your branch you are good to merge. 

What about docs? should you update the info where this command imentioned?

mdelez:
> 

The `yalc-publish` script that I renamed as `yalc-publish-and-update` has the `--push` flag which will automatically update any local repositories using the JS-LIB. This isn't an issue with the e2e tests because the setup is confined. But it can become an annoyance when running the script locally because it will update, most notably, the JS-LIB version in your local DSP-APP repo and you have to remember to not commit this change and can also cause confusion if you're working on two unrelated tasks in the JS-LIB and DSP-APP because you could update your JS-LIB version without realizing. In any case, the steps in the docs are still correct but I added a part about the new npm script in [399020e](https://github.com/dasch-swiss/dsp-js-lib/pull/504/commits/399020e7282b11187e8cd89df0c72ddbd4f4b3ed)

PR #547, opened 2023-08-02
feat(search-endpoint): add method for search by label count

mdelez:
resolves DEV-2507

### DIFF
--- a/test-framework/src/app/app.component.ts
+++ b/test-framework/src/app/app.component.ts
@@ -1460,6 +1460,16 @@ export class AppComponent implements OnInit {
         );
     }
 
+    labelSearchCount(searchTerm: string) {
+        
+        this.knoraApiConnection.v2.search.doSearchByLabelCountQuery(searchTerm).subscribe(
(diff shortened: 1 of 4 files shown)


test-framework/src/app/app.component.ts
```diff
@@ -1460,6 +1460,16 @@ export class AppComponent implements OnInit {
         );
     }
 
+    labelSearchCount(searchTerm: string) {
+        
+        this.knoraApiConnection.v2.search.doSearchByLabelCountQuery(searchTerm).subscribe(
+            (res: CountQueryResponse) => {
+                console.log(res);
```
Vijeinath, line 1467: delete console.log

Vijeinath (approved):

test-framework/src/app/app.component.ts
mdelez: it's a bit weird but it's useful to have console.logs in the test app so you can be sure something happened when you click the corresponding button in the page of buttons that is our test app

PR #639, opened 2024-05-21
feat!: Send X-Asset-Ingested header for update/create of value and resources (DEV-2628)

seakayone:
This represents a breaking change in the sense that when using the dsp-js-lib the asset upload must be done using dsp-ingest and cannot be done using Sipi anymore.


irmastnt (approved):
Ok for me, just, as a side note using constants would be beneficial in order to decrease repetitive code.

seakayone:
> Ok for me, just, as a side note using constants would be beneficial in order to decrease repetitive code.

Good point. I would prefer that as well. Unfortunately I have not been able to add it to the `Constants.ts`, import and use it as the key for the headerOpts. Keeps complaining that the `string` is not a `RecursiveString` something.

Next time maybe?

PR #649, opened 2025-03-06
feat: endpoints values do not return ApiResponseError type 

derschnee68:
resolves DEV-

### DIFF
--- a/src/cache/ListAdminCache.spec.ts
+++ b/src/cache/ListAdminCache.spec.ts
@@ -5,6 +5,7 @@ import { KnoraApiConfig } from "../knora-api-config";
 import { KnoraApiConnection } from "../knora-api-connection";
 import { ListResponse } from "../models/admin/list-response";
 import { ApiResponseData } from "../models/api-response-data";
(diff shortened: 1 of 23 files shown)


src/cache/ListAdminCache.spec.ts
```diff
@@ -34,7 +35,7 @@ describe("ListCache", () => {
         getListSpy = spyOn(knoraApiConnection.admin.listsEndpoint, "getList").and.callFake(
             (listIri: string) => {
 
-                return of({body: listResp} as ApiResponseData<ListResponse> | ApiResponseError);
+                return of({body: listResp} as ApiResponseData<ListResponse> | ApiResponseData<ListChildNodeResponse>);
```
mpro7, line 38: In all other places you have removed strong typing, but here something is added. It looks fine as a first glance, but I van't verify it now. Just double check if this is the desire type returned here. There are small changes between parent list and and children.

mpro7 (approved):
Looks good, check my comment below.

